Issue #44, opened 2023-04-21
Full send_buf prevents sending of FIN


njgheorghita:
I've spent some time playing around with this library. Basically, I've been running a bare simple app locally that creates a sender & receiver utp socket and tries to tx 100000 bytes. This transfer has failed consistently - while it might be user error, i'm fairly confident I'm using the library in the way it's intended...

The repeated behavior I'm seeing is the sender / initiator processes the tx succesfully and closes the stream - however the recipient never fully processes the stream (via `read_to_eof`) and hangs until it time outs. 

It appears to me as though a `fin` is never sent from the sender -> receipient. This is because the sender never clears its `send_buf`. The evaluations [here](https://github.com/jacobkaufmann/utp/blob/master/src/conn.rs#L329) and [here](https://github.com/jacobkaufmann/utp/blob/master/src/conn.rs#L377) **always** fail since `send_buf` is never emptied. So the fin is never sent. Something appears to be happening [here](https://github.com/jacobkaufmann/utp/blob/master/src/conn.rs#L439-L467) where the `send_buf` is not properly cleared.

I still have some understanding to improve upon, so I'm not sure if these thoughts make any sense. 
- i haven't grokked the supposed behavior b/w `send_buf` and `pending_writes`
  - while the `pending_writes` queue seems to empty correctly, the `send_buf` never fully clears... porque?
- why is `send_buf` a `VecDeque<Vec<u8>>` and not just `VecDeque<u8>>`? 

But from what I can tell from experimenting with this library, it's consistent behavior that's preventing the sending of the `fin` packet. Mostly just leaving this here so that I can try and organize my thoughts & i'll pick it back up after the weekend,  but if anybody has any clues / ideas / pointers on what I'm misunderstanding - that's definitely appreciated.

jacobkaufmann:
can you share calling code to reproduce?

njgheorghita:
:+1: on send buffer

Yeah, I also substituted 100k into that test and everything worked. I was running with `rt-multi-thread` active and experienced the same behavior after changing `[tokio::main]` -> `#[tokio::main(flavor = "multi_thread", worker_threads = 10)]`.

In #45 I explained some weird behavior in my terminal that I observed. And since in that pr we have a single test to verify against, which is easier to debug, I'll revisit this issue once we resolve whatever's going on in that pr.

emhane:
[...]

PR #54, opened 2023-05-23
Debug stream

emhane:
On my MacBook Air M1 I can transfer 1 million bytes no problem. When I try to transfer 2 million bytes I get this error when running on my branch _error-handling_ `thread 'stream::test::test_transfer_2m_bytes' panicked at 'Should send 1m bytes: Conn(ConnClosed("idle uTP connection time out"))'`. This closes https://github.com/ethereum/utp/issues/44, possibly in place of a new issue.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -165,7 +165,7 @@ mod test {
         let config = ConnectionConfig::default();
 
         // write 100k bytes data to the remote peer over the stream.
-        let data = vec![0xed; 1000_000];
+        let data = vec![0xed; 1_000_000];
         let rx_handle = tokio::spawn(async move {
             receiver.accept(config).await.unwrap();
         });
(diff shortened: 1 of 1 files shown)


KolbyML:
mine fails at exactly 1,100,000.

KolbyML:
image
^ this is running 1,100,000 and the code error's out
image
^ this is running anything below that number.

So it seems like if the number is too high 0 packets are sent at all

KolbyML:
Ok the bug appears to be that when a vector is too big the payload creation code creates 0 payloads

KolbyML:
Ok I found the bug I will make a PR with a patch

KolbyML:
I wouldn't blame this on hardware, the software is just affected by the hardware which should be considered a side affect of the code and use of generics

PR #56, opened 2023-05-23
Solution 1# for client halt bug

KolbyML:
fixes https://github.com/ethereum/utp/pull/54 and https://github.com/ethereum/utp/issues/44

### The issue was / how to replicate the bug
```rust
        // Write as much data as possible into send buffer.
        while let Some((data, ..)) = self.pending_writes.front() {
            if data.len() <= send_buf.available() {
                let (data, tx) = self.pending_writes.pop_front().unwrap();
                send_buf.write(&data).unwrap();
                let _ = tx.send(Ok(data.len()));
                self.writable.notify_one();
            } else {
                break;
            }
    }
```

If data.len() is > send_buf.available then the code will "lock"/stale data from being able to be sent. A.k.a data will never be added to the send buffer resulting in a timeout with 0 packets sent.

pictures of wireshark testing this bug can be seen here https://github.com/ethereum/utp/pull/54#issuecomment-1559872700

### where did this bug come from?
I believe it is because receive buffer was written first where having a limit makes sense (for sending the final packet), but then the code was copy pasted creating send buffer this limit was kept but didn't have a usecase, but still had checks which then caused the stall.

### How was the bug fixed?
By removing avaliability/size limit  from send buffer and hence the unneeded checks. Resolving the stall send bug.

To iterate there isn't a reason why we should have that limit for send buffer. There is already code that handles windowing packets.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -70,14 +70,14 @@ enum State<const N: usize> {
     Connecting(Option<oneshot::Sender<io::Result<()>>>),
     Established {
         recv_buf: ReceiveBuffer<N>,
-        send_buf: SendBuffer<N>,
+        send_buf: SendBuffer,
         sent_packets: SentPackets,
     },
     Closing {
         local_fin: Option<u16>,
         remote_fin: Option<u16>,
         recv_buf: ReceiveBuffer<N>,
-        send_buf: SendBuffer<N>,
+        send_buf: SendBuffer,
         sent_packets: SentPackets,
     },
     Closed {
@@ -455,15 +455,11 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         }
 
         // Write as much data as possible into send buffer.
-        while let Some((data, ..)) = self.pending_writes.front() {
-            if data.len() <= send_buf.available() {
-                let (data, tx) = self.pending_writes.pop_front().unwrap();
-                send_buf.write(&data).unwrap();
-                let _ = tx.send(Ok(data.len()));
-                self.writable.notify_one();
-            } else {
-                break;
-            }
+        while self.pending_writes.front().is_some() {
+            let (data, tx) = self.pending_writes.pop_front().unwrap();
+            send_buf.write(&data).unwrap();
+            let _ = tx.send(Ok(data.len()));
+            self.writable.notify_one();
         }
 
         // Transmit data packets.
--- a/src/send.rs
+++ b/src/send.rs
@@ -4,32 +4,24 @@ use std::io;
 type Bytes = Vec<u8>;
 
 #[derive(Clone, Debug)]
-pub struct SendBuffer<const N: usize> {
+pub struct SendBuffer {
     pending: VecDeque<Bytes>,
     offset: usize,
 }
 
-impl<const N: usize> Default for SendBuffer<N> {
+impl Default for SendBuffer {
     fn default() -> Self {
-        Self {
-            pending: VecDeque::new(),
-            offset: 0,
-        }
+        Self::new()
     }
 }
 
-impl<const N: usize> SendBuffer<N> {
+impl SendBuffer {
     /// Creates a new buffer.
     pub fn new() -> Self {
         Self {
             pending: VecDeque::new(),
             offset: 0,
         }
-    }
-
-    /// Returns the number of bytes available in the buffer.
-    pub fn available(&self) -> usize {
-        N - self.pending.iter().fold(0, |acc, x| acc + x.len()) + self.offset
     }
 
     /// Returns `true` if the buffer is empty.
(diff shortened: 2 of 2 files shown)


KolbyML:
@jacobkaufmann @jacobkaufmann @emhane

ping for review

emhane:
I pulled your fix, I can send at least 10 MB no problem now.

Could you link the code lines you base your assumptions on? It helps to argue for your case by speeding up the process for readers tying to recreate your trail of thought. I'm interested in "There is already code that handles windowing packets", because many times there are reasons to mange congestion control on multiple levels.

Rn multiplexing is like this: 
one udp socket to one utp socket
one utp socket to many streams
one stream to one connection
one connection to one congestion control

One congestion controller per stream exists (conn [acceptor](https://github.com/ethereum/utp/blob/a2b619d2b48df347af7ff305d8022caf887868a1/src/conn.rs#L213-L220) and [initiator](https://github.com/ethereum/utp/blob/a2b619d2b48df347af7ff305d8022caf887868a1/src/conn.rs#L800-L801)), hence it seems the congestion controller is accounting for the concurrent sending and receiving of packets on that stream,  but isn't managing congestion control of concurrent streams, at least not in an informed way since that would require the congestion controller to know the state of the other congestion controllers (other streams). We only have one udp socket to receive and send packets from many streams on and we want those streams to progress concurrently, so congestion control of streams does matter. Therefore we may need the check `if data.len() <= send_buf.available() {` after all for this purpose. 

Nonetheless, this fix doesn't let me send 100 MB, which it should to be stable. Running your fix on my _error-handling_ branch I get this informative error: `thread 'stream::test::test_transfer_100_megabyte' panicked at 'Should read 100 megabyte: Conn(ConnClosed("idle uTP connection time out"))'`, which leads me to think the error originates somewhere [here where the idle time out is reset](https://github.com/ethereum/utp/blob/a2b619d2b48df347af7ff305d8022caf887868a1/src/conn.rs#L266-L272). Adding the connection endpoint to that error tells me it is the acceptor of the connection that times out which in my test is only receives data: `thread 'stream::test::test_transfer_100_megabyte' panicked at 'Should read 100 megabyte: Conn(ConnClosed("idle uTP connection time out, Acceptor((10328, 10650))"))'`. 

Running this test for 10 MB without your fix Kolby, I get the same idle time out error, but from the initiator side, which in my test only sends data: `thread 'stream::test::test_transfer_10_megabyte' panicked at 'Should send 100 megabyte: Conn(ConnClosed("idle uTP connection time out, Initiator((36072, 1))"))`.

By this information, I think that use of the idle time out is funky and needs to be inspected.

Great work!

KolbyML:
:shipit: good dective work I will debug with 100MB now. Recieveing a packet should reset the idle time in theory so maybe I will look at that

KolbyML:
> Could you link the code lines you base your assumptions on? It helps to argue for your case by speeding up the process for readers tying to recreate your trail of thought. I'm interested in "There is already code that handles windowing packets", because many times there are reasons to mange congestion control on multiple levels.

It would probably just be quicker to control f send_buffer then control f receive_buffer. The limited shouldn't affect congestion controls but I guess I should look into it more.

KolbyML:
100mb fail is caused by a timeout bug, cause when I run wire shark the data is being sent fine.

So this PR fixes a stale bug, I will try to find the timeout bug

jacobkaufmann:
this is not a bug, and the limit on the send buffer is deliberate. there is a [`TODO`](https://github.com/ethereum/utp/blob/a2b619d2b48df347af7ff305d8022caf887868a1/src/stream.rs#L12) to make the capacity configurable.

KolbyML:
> this is not a bug, and the limit on the send buffer is deliberate. there is a [`TODO`](https://github.com/ethereum/utp/blob/a2b619d2b48df347af7ff305d8022caf887868a1/src/stream.rs#L12) to make the capacity configurable.

I am not saying having a limit is a bug. I am saying 
```rust
        // Write as much data as possible into send buffer.
        while let Some((data, ..)) = self.pending_writes.front() {
            if data.len() <= send_buf.available() {
                let (data, tx) = self.pending_writes.pop_front().unwrap();
                send_buf.write(&data).unwrap();
                let _ = tx.send(Ok(data.len()));
                self.writable.notify_one();
            } else {
                break;
            }
    }
```

If this else condition is hit the client will stall which doesn't seem to be intentional? Hence the stall is unintential? 

image
^ when the else condition is hit this is what happens
image
^ this is what is supposed to happen

So the limit it itself isn't the problem, it is how it is handled.

jacobkaufmann:
yea sorry I think I misinterpreted your comment. what I think should happen is that an attempt to write returns an error when the send buffer (or the sum of send buffer and pending writes) exceeds the capacity. the capacity exists to limit the amount of resource consumption. related, the capacity should be set to a value that is greater than or equal to the largest conceivable write your program will perform.

the separation between pending writes and send buffer exists so that send buffer can be quite simple. I'm not saying a simpler design isn't available, but the solution here where the capacity of the send buffer is removed is not in the right direction.

KolbyML:
> the separation between pending writes and send buffer exists so that send buffer can be quite simple. I'm not saying a simpler design isn't available, but the solution here where the capacity of the send buffer is removed is not in the right direction.

I think that is fair, so I think the short term solution would then be to return an error like you said then add a todo possibly to simply that

KolbyML:
closed in favor of https://github.com/ethereum/utp/pull/61

PR #59, opened 2023-05-24
Add else condition for writing as much into buffer alternate solution to https://github.com/ethereum/utp/pull/56

KolbyML:
An alternative solution to https://github.com/ethereum/utp/pull/56

This solves the halt problem, by pushing as much as we can and basically pushing the bytes sent and handing the problem to the user of the library. 

This will make it so instead of the library handing the issue the user will have to retry at a later date

This PR's solution is flawed reasoning stated here https://github.com/ethereum/utp/pull/56#issuecomment-1561668474


KolbyML:
closed in favor of https://github.com/ethereum/utp/pull/61

Issue #57, opened 2023-05-24
UTP Interop 


njgheorghita:
More of a comment than an actual issue... But I think this could be of value to those of us who are debugging this library atm.

I wanted to confirm my suspicion as to what was causing the difference in performance when running glados against fluffy / trin. For context, running glados against fluffy resulted in drastically improved performance when compared to running glados against trin. It appeared as though fluffy was having no trouble with utp transfers, while trin was struggling. Since `trin-bridge` is the only bridge running, the conclusion was that utp txs for `trin` -> `fluffy` were working just fine, whereas `trin` -> `trin` was broken (for bodies & receipts ... aka any significantly long utp tx).

Results from running `trin-bridge` in backfill mode for 15mins. `trin-bridge` is run in isolation against a single client. Each utp tx is logged and marked as success & failure.

# trin-bridge -> trin
<img width="777" alt="Screen Shot 2023-05-11 at 3 36 49 PM" src="https://github.com/ethereum/utp/assets/9753150/86b0fd0e-bdb6-4e0b-9614-7c9e97f3db49">

# trin-bridge -> fluffy
<img width="842" alt="Screen Shot 2023-05-24 at 12 05 40 PM" src="https://github.com/ethereum/utp/assets/9753150/50896d61-d2e1-4574-afc8-bc95b9bcc2bc">

For whatever reason, it appears as though `trin` -> `fluffy` utp txs work flawlessly. Whereas `trin` -> `trin` utp txs are problematic. Hopefully this can help shed some insight as to where the bugs might reside.

PR #60, opened 2023-05-25
2 tests demonstrating the overflow bug

KolbyML:
Details of how I found the bug and information are detailed in a comment linked here https://github.com/ethereum/utp/issues/44#issuecomment-1563475271

seq_nr/ack_nr are both u16 this is intential TCP/uTP have intentional overflows so that they can continuously send data

In this demo pr test case test_transfer_65535_packets_in_theory_should_pass sends 65535 1 below the limit and passes

but when the overflow occurs test_transfer_65536_packets_in_theory_should_fail it fails

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -119,3 +119,101 @@ impl<P> Drop for UtpStream<P> {
         let _ = self.shutdown();
     }
 }
+
+#[cfg(test)]
+mod test {
+    use crate::conn::ConnectionConfig;
+    use crate::socket::UtpSocket;
+    use std::net::SocketAddr;
+
+    #[tokio::test]
+    async fn test_transfer_65535_packets_in_theory_should_pass() {
+        // set-up test
+        tracing_subscriber::fmt::init();
+        let sender_addr = SocketAddr::from(([127, 0, 0, 1], 3700));
+        let receiver_addr = SocketAddr::from(([127, 0, 0, 1], 3701));
+
+        let sender = UtpSocket::bind(sender_addr).await.unwrap();
+        let receiver = UtpSocket::bind(receiver_addr).await.unwrap();
+
+        let config = ConnectionConfig::default();
+
+        // accept connection
+        let rx = async move {
+            let mut rx_stream = receiver.accept(config).await.unwrap();
+            // read data from the remote peer until the peer indicates there is no data left to
+            // write.
+            let mut data = vec![];
+            rx_stream
+                .read_to_eof(&mut data)
+                .await
+                .expect("Should read 1 megabyte")
+        };
+
+        let tx = async move {
+            // write 100k bytes data to the remote peer over the stream.
+            let data = vec![0xef; 1];
+            let mut tx_stream = sender.connect(receiver_addr, config).await.unwrap();
+
+            let mut letter = 0;
+            for _ in  1..65535 {
+                letter += tx_stream
+                    .write(data.as_slice())
+                    .await
+                    .expect("Should send 1 megabyte");
+            }
+
+            letter
+        };
+
+        let (tx_res, rx_res) = tokio::join!(tx, rx);
+
+        assert_eq!(tx_res, rx_res);
+    }
+
+    #[tokio::test]
+    async fn test_transfer_65536_packets_in_theory_should_fail() {
+        // set-up test
+        tracing_subscriber::fmt::init();
+        let sender_addr = SocketAddr::from(([127, 0, 0, 1], 3800));
+        let receiver_addr = SocketAddr::from(([127, 0, 0, 1], 3801));
+
+        let sender = UtpSocket::bind(sender_addr).await.unwrap();
+        let receiver = UtpSocket::bind(receiver_addr).await.unwrap();
+
+        let config = ConnectionConfig::default();
+
+        // accept connection
+        let rx = async move {
+            let mut rx_stream = receiver.accept(config).await.unwrap();
+            // read data from the remote peer until the peer indicates there is no data left to
+            // write.
+            let mut data = vec![];
+            rx_stream
+                .read_to_eof(&mut data)
+                .await
+                .expect("Should pass to send 65535 packets")
+        };
+
+        let tx = async move {
+            // write 100k bytes data to the remote peer over the stream.
+            let data = vec![0xef; 1];
+            let mut tx_stream = sender.connect(receiver_addr, config).await.unwrap();
+
+            let mut letter = 0;
+            for _ in  1..65536 {
+                letter += tx_stream
+                    .write(data.as_slice())
+                    .await
+                    .expect("Should fail to send 65536 packets");
+            }
+
+            letter
+        };
+
+        let (tx_res, rx_res) = tokio::join!(tx, rx);
+
+        assert_eq!(tx_res, rx_res);
+    }
+}
+


KolbyML:
image
@emhane @jacobkaufmann @njgheorghita

Here is a picture of the test log of the demo working as expected :partying_face:

KolbyML:
Looking into it a little there seems to be a few issues. Having ``init_seq_num`` isn't very useful instead it should be replaced with something like ``conn.seq``  like libutp does. ``init_seq_num`` seems like a precursor to a bug like this occuring since it assumes we won't reach the number again (which we will)

Instead we should keep track of the`` sequence`` and a ``current_window_packets`` (just a count) this is the way it is done in fluffy's uTP and the reference implementation. This way we can track if ack's relatively instead of based off a fixed point, because we do reuse numbers TCP does it too.

I am not done researching for a solution to this problem, but this kind of seems like a starting point. ``let range = self.seq_num_range();`` seems buggy assuming there is only u16 packets ever sent kind of thing. When infinity packets can be sent so knowing the initial seq isn't helpful knowing the sequence and how many packets are in the window are what matter since everything is relative.

I guess I will test more and see what I come up with

jacobkaufmann:
in your screenshot it looks like the test that was supposed to pass is not terminating. but yea `init_seq_num` would present a problem for a connection that transmits more than `u16::MAX` packets because of the `seq_num_range` method.

I did not anticipate that it would be a problem for its initial use cases. if each `DATA` packet contains ~500-1000 bytes of data, then you would have to send on the order of tens of megabytes of data to run into that issue. I'm not saying it's not an issue, but I would not expect that issue to come up in practice unless you are doing large (tens of megabytes) data transfers.

KolbyML:
> in your screenshot it looks like the test that was supposed to pass is not terminating.

``test stream::test::test_transfer_65535_packets_in_theory_should_pass ... ok`` ? it did terminate

> but yea `init_seq_num` would present a problem for a connection that transmits more than `u16::MAX` packets because of the `seq_num_range` method. I did not anticipate that it would be a problem for its initial use cases. if each `DATA` packet contains ~500-1000 bytes of data, then you would have to send on the order of tens of megabytes of data to run into that issue. I'm not saying it's not an issue, but I would not expect that issue to come up in practice unless you are doing large (tens of megabytes) data transfers.

I do believe this issue would currently not come up in something like the history network from the data I have heard from nick.

In the most common usecase of uTP bittorent (uTP is quite heavily used from what I have seen using qTorrent) it is not uncommon to see above 65 megabytes transfers. A benefit of trin is the transfers should be low, but if it is used like a TCP socket which it is basically a clone of it probably will reach that. 

It is a public utp implementation and currently the only one now maintained on crates.io so I think this is worth it to fix.

Anyways I will look into making a PR to resolve this, I am just trying to find/fix bugs in the library cause as seen in https://github.com/ethereum/utp/issues/57 it is quite an important issue. I believe for our current usecase the @emhane solution will probably bear more fruit, but it is best IMO to just get all the underlying problems out of the way.

emhane:
well done @KolbyML this is spot on. no more than 65535 packets can be sent due to the seq num being a u16, which in fact, is just a megabyte of data. I was also wondering if we could either 1: adapt the protocol for ethereum needs making the seq num a larger number or 2: rotate the seq num, as you mention in #62 .

emhane:
> > in your screenshot it looks like the test that was supposed to pass is not terminating.
> 
> `test stream::test::test_transfer_65535_packets_in_theory_should_pass ... ok` ? it did terminate
> 
> > but yea `init_seq_num` would present a problem for a connection that transmits more than `u16::MAX` packets because of the `seq_num_range` method. I did not anticipate that it would be a problem for its initial use cases. if each `DATA` packet contains ~500-1000 bytes of data, then you would have to send on the order of tens of megabytes of data to run into that issue. I'm not saying it's not an issue, but I would not expect that issue to come up in practice unless you are doing large (tens of megabytes) data transfers.
> 
> From what I know this issue shouldn't affect the history network currently. Though it is the only public utp implementation on crates.io currently being maintained. So I think having a stable library would be beneficial. uTP is just a TCP clone with a "fancier "congestion protocol" on UDP. When I use qBittorrent over half of my connections are often uTP instead of TCP, transfers are often over that possible 32-65mb failure range. So I think that is kind of the feature of uTP to do large transfers of data over 1kb and beyond.
> 
> I am just trying to catch them all, as the saying goes. The bug @emhane is tracking will probably bear more fruit for history. Hopefully the collective work though can get this library as stable as libtorrent 💪
> 
> I will make a PR resolving this issue

I totally agree with your argument @KolbyML. Good observations on the weight of a rust uTP crate in the current time, and the public expectations on how much data the uTP protocol should handle. Indeed, since the other rust uTP crates aren't maintained anymore, it makes the interoperability of this crate with for example libtorrent important. Nice to have a number for common max transfer size for uTP, 32-65 MB, that helps us set the milestone towards which to work. If you have some links or self generated data to show the failure range is 32-65 MB from other non-rust uTP implementations, please share.

KolbyML:
> well done @KolbyML this is spot on. no more than 65535 packets can be sent due to the seq num being a u16, which in fact, is just a megabyte of data. I was also wondering if we could either 1: adapt the protocol for ethereum needs making the seq num a larger number or 2: rotate the seq num, as you mention in #62 .

2 is following the spec, 1 wouldn't be uTP

KolbyML:
> I totally agree with your argument @KolbyML. Good observations on the weight of a rust uTP crate in the current time, and the public expectations on how much data the uTP protocol should handle. Indeed, since the other rust uTP crates aren't maintained anymore, it makes the interoperability of this crate with for example libtorrent important. Nice to have a number for common max transfer size for uTP, 32-65 MB, that helps us set the milestone towards which to work. If you have some links or self generated data to show the failure range is 32-65 MB from other non-rust uTP implementations, please share.


32-65MB is just calculated from @jacobkaufmann range of 500-1000 bytes. 500*65535 = 32MB.

Other uTP implementation's don't have this failure, this is only a failure that happens with our client which is what makes it a bug.


### TL:DR

Every other compliant uTP implementation can send infinity amount of data

Our implementation can only send 65535 packets of data before failing because of a bug in implementation design

Sorry if my original message was confusing, but this is a ethereum/utp only bug.

emhane:
> > I totally agree with your argument @KolbyML. Good observations on the weight of a rust uTP crate in the current time, and the public expectations on how much data the uTP protocol should handle. Indeed, since the other rust uTP crates aren't maintained anymore, it makes the interoperability of this crate with for example libtorrent important. Nice to have a number for common max transfer size for uTP, 32-65 MB, that helps us set the milestone towards which to work. If you have some links or self generated data to show the failure range is 32-65 MB from other non-rust uTP implementations, please share.
> 
> 32-65MB is just calculated from @jacobkaufmann range of 500-1000 bytes. 500*65535 = 32MB.
> 
> Other uTP implementation's don't have this failure, this is only a failure that happens with our client which is what makes it a bug.
> 
> ### TL:DR
> Every other compliant uTP implementation can send infinity amount of data
> 
> Our implementation can only send 65535 packets of data before failing because of a bug in implementation design
> 
> Sorry if my original message was confusing, but this is a ethereum/utp only bug.

gotcha, then the goal is to transfer infinitely many bytes.

the limit to the amount of data we can transfer on a stream rn on master is 1 MB with 65535 ACKs. so not even close to 32-65 MB.

KolbyML:
> gotcha, then the goal is to transfer infinitely many bytes.

yeppers

> the limit to the amount of data we can transfer on a stream rn on master is 1 MB with 65535 ACKs. so not even close to 32-65 MB.

The send buffer limit (the 1 MB you are talking about) isn't 1 MB though, if you fill pending_sends with 0.9MB you can send over 1MB per a stream and exceed 32-65MB.

KolbyML:
> gotcha, then the goal is to transfer infinitely many bytes.

yeppers

> the limit to the amount of data we can transfer on a stream rn on master is 1 MB with 65535 ACKs. so not even close to 32-65 MB.

The send buffer limit (the 1 MB you are talking about) doesn't mean we can only send 1 MB per a stream, if you fill pending_sends with 0.9MB you can send over 1MB per a stream and exceed 32-65MB.

PR #61, opened 2023-05-26
Prevent client from deadlocking/halting on large transfer by throwing error

KolbyML:
The issue can be read in this PR https://github.com/ethereum/utp/pull/56

This is a quick solution to the problem, with a further improvement added as a TODO

ping for request to review @jacobkaufmann


KolbyML:
Closing since I think we should merge https://github.com/ethereum/utp/pull/65 as the temporary solution to solve this problem. It adds complexity, but we will have to chunk data either way. Hopefully soon we can refactor the two pending write buffers into 1

PR #62, opened 2023-05-28
Refactor to fix overflow bug part 1

KolbyML:
information about the bug can be found here https://github.com/ethereum/utp/pull/60

I decided I will make this fix into a few parts to make it easier to review changes.

I implemented ``SizableCircularBuffer`` based off reference implementation bittorrent/libutp. Currently I have it implemented so it will be backwards compatible with the previous implementation for now till the next part. I did this to make it easier to review. There are comments signifying the compatibility code to be removed in the future. This data structure will be used for the fix in the next parts.

Also in ``SentPackets`` I removed ``init_seq_num`` and replaced it with ``seq_num`` this will actively keep track of the current ``seq_num``. This is based off the reference implementation and is needed for the fix.

### Plans for the next part

I will look to implement something like ``cur_window_packets`` which is a count of the packets that are sent but currently not acked. I will also implement ``ack_num`` tracking cause currently the way we find the currently acted packet won't be able to work with the plan for the patch according to my research. This part may be split into 2 depending on how big the changes end up for reviewability.

### Final notes
Hopefully we can get this fixed within 2 PR's but we will see. Trying to optimize for reviewability.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,3 +11,4 @@ pub mod socket;
 pub mod stream;
 pub mod time;
 pub mod udp;
+pub mod utils;
--- a/src/sent.rs
+++ b/src/sent.rs
@@ -4,6 +4,7 @@ use std::time::{Duration, Instant};
 use crate::congestion;
 use crate::packet::{PacketType, SelectiveAck};
 use crate::seq::CircularRangeInclusive;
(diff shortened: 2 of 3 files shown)


KolbyML:
@jacobkaufmann @jacobkaufmann @emhane ping for review

PR #63, opened 2023-05-28
Fix FIN not sent bug

emhane:
Closes https://github.com/ethereum/utp/issues/44.

A read doesn't complete until a write completes. A write doesn't complete unless a FIN is sent. A FIN is sent when the stream shuts down, hence it should be included in the `write` public function. Furthermore, one stream maps to one connection and therefore it makes no sense to keep the the reading stream open when the write stream is shut down, hence I have included shutdown in `read_to_eof` too.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -66,24 +66,22 @@ where
                 .send((buf.capacity(), tx))
                 .map_err(|_| io::Error::from(io::ErrorKind::NotConnected))?;
 
-            match rx.await {
-                Ok(result) => match result {
-                    Ok(mut data) => {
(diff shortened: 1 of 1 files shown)


emhane:
possibly also closes https://github.com/ethereum/utp/pull/45

KolbyML:
> possibly also closes #45

This doesn't close #45 fixing the overflow bug would close that.

https://github.com/ethereum/utp/pull/45#issuecomment-1538768973

Where he explains how he is testing, the reason these cases are failing are do it sending 2^16 packets aka the "overflow bug" I am trying to fix

PR #65, opened 2023-05-28
Avoid deadlock on data bigger than buffer by chunking

emhane:
A deadlock occurs when the data to send is bigger than the available buffer. Chunking data removes this deadlock.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -455,15 +455,20 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         }
 
         // Write as much data as possible into send buffer.
-        while let Some((data, ..)) = self.pending_writes.front() {
-            if data.len() <= send_buf.available() {
-                let (data, tx) = self.pending_writes.pop_front().unwrap();
-                send_buf.write(&data).unwrap();
-                let _ = tx.send(Ok(data.len()));
-                self.writable.notify_one();
-            } else {
+        while send_buf.available() > 0 {
+            let Some((data, tx)) = self.pending_writes.pop_front() else {
                 break;
+            };
+            let written = send_buf.write(&data).unwrap();
+            if written < data.len() {
+                // not all data fit in the send buffer, chunk data
+                let mut data = data;
+                let remaining = data.split_off(send_buf.available());
+                self.pending_writes.push_front((remaining, tx));
+            } else {
+                let _ = tx.send(Ok(data.len()));
             }
+            self.writable.notify_one();
         }
 
         // Transmit data packets.


KolbyML:
For tracking this is an alternative to https://github.com/ethereum/utp/pull/61

src/conn.rs
```diff
@@ -455,15 +455,20 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
...
+            let written = send_buf.write(&data).unwrap();
+            if written < data.len() {
+                // not all data fit in the send buffer, chunk data
+                let mut data = data;
+                let remaining = data.split_off(send_buf.available());
+                self.pending_writes.push_front((remaining, tx));
+            } else {
+                let _ = tx.send(Ok(data.len()));
             }
+            self.writable.notify_one();
```
KolbyML, line 471: Isn't the issue with this implementation ``let _ = tx.send(Ok(data.len()));`` when this returns that the data was sent to the user, the ``data.len()`` will be wrong. This is a possible solution I thought of, but to fix the issue I listed above would take keeping track off how much data was written per channel which I wasn't a fan of.

emhane: not quite sure what you mean here? I caught a bug by having a second look at it, didn't store the value of `send_buf.available()` then modified it, then used it again

KolbyML: If your transfer is split into 2 5MB 5MB but the total is 10MB

Won't let _ = tx.send(Ok(data.len())); now send 5MB when it was supposed to be 10MB returned

emhane: oh yeah, that's right! thanks!

KolbyML: This is why I made the PR for this solution https://github.com/ethereum/utp/pull/61 . Cause fixing the problem at a deeper level requires a refactor, unless you increase complexity. Which I guess is fine temporary.

The refactor would be  something like combining pending_writes and send_buf into one, since having them both is quite redundant since we already have an outgoing buffer which is ``sent_packets.packets`` ``packets`` should be renamed to ``outgoing_buffer`` to better represent the reference implementations

KolbyML:
https://github.com/ethereum/utp/pull/65#discussion_r1208613588

This is why I made the PR for this solution https://github.com/ethereum/utp/pull/61 . Cause fixing the problem at a deeper level requires a refactor, unless you increase complexity. Which I guess is fine temporary.

The refactor would be something like combining pending_writes and send_buf into one, since having them both is quite redundant since we already have an outgoing buffer which is sent_packets.packets packets should be renamed to outgoing_buffer to better represent the reference implementations

KolbyML:
If we choose your temporary solution over mine, you should add a TODO to this PR of.

``// todo: combine pending_writes and send_buf, rename sent_packets.packets to sent_packets.outgoing_buffer``

PR #66, opened 2023-05-28
Not panic on pending data larger than buffer

emhane:
There is no need to panic when more values are pending than the buffer capacity

### DIFF
--- a/src/recv.rs
+++ b/src/recv.rs
@@ -28,10 +28,8 @@ impl<const N: usize> ReceiveBuffer<N> {
 
     /// Returns the number of available bytes remaining in the buffer.
     pub fn available(&self) -> usize {
-        N.saturating_sub(
-            self.offset
-                .saturating_sub(self.pending.values().fold(0, |acc, data| acc + data.len())),
-        )
+        N.saturating_sub(self.offset)
+            .saturating_sub(self.pending.values().fold(0, |acc, data| acc + data.len()))
     }
 
     /// Returns `true` if the buffer is empty.
--- a/src/recv.rs
+++ b/src/recv.rs
@@ -28,7 +28,10 @@ impl<const N: usize> ReceiveBuffer<N> {
 
     /// Returns the number of available bytes remaining in the buffer.
     pub fn available(&self) -> usize {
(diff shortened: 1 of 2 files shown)


src/recv.rs
```diff
@@ -28,7 +28,8 @@ impl<const N: usize> ReceiveBuffer<N> {
 
     /// Returns the number of available bytes remaining in the buffer.
     pub fn available(&self) -> usize {
-        N - self.offset - self.pending.values().fold(0, |acc, data| acc + data.len())
+        N.saturating_sub(self.offset)
+            .saturating_sub(self.pending.values().fold(0, |acc, data| acc + data.len()))
```
KolbyML, line 32:
```suggestion
        N  + self.offset - self.pending.values().fold(0, |acc, data| acc + data.len())
```

Try this I believe the issue is, data is added to pending, then when the data is read offset increases which causes the overflow

So we can either remove offset or add offset to N

src/send.rs
```diff
@@ -29,7 +29,7 @@ impl<const N: usize> SendBuffer<N> {
 
     /// Returns the number of bytes available in the buffer.
     pub fn available(&self) -> usize {
-        N - self.pending.iter().fold(0, |acc, x| acc + x.len()) + self.offset
+        N.saturating_sub(self.pending.iter().fold(0, |acc, x| acc + x.len()) + self.offset)
```
KolbyML, line 32: This should already be fine, the issue with recv.rs is that it minus's the offset.

if you want you can try
`` N - self.pending.iter().fold(0, |acc, x| acc + x.len()) + self.offset``

emhane: I see I messed up those brackets

KolbyML: I don't think we should use saturing_sub/add since if there is a bug in the code this will hide it. If there is no order bug then this should never panic

src/recv.rs
KolbyML: Remember ``N.saturating_sub(self.offset)`` should be add not sub, but we probably shouldn't use saturing since it will hide bugs in the underlying logic

KolbyML:
Closed with https://github.com/ethereum/utp/pull/77

PR #70, opened 2023-06-05
limit congestion control timeout amplification

jacobkaufmann:
we have observed that the congestion control timeout can amplify quickly to the point where the timeout becomes not so useful. for example, the timeout should not exceed the maximum idle timeout.

summary:

* add a maximum timeout field to congestion control
* limit the number of timeouts registered with congestion control by the connection
* add a maximum timeout field to the connection config
  * by default, this value is equal to the max idle timeout

### DIFF
--- a/src/congestion.rs
+++ b/src/congestion.rs
@@ -687,12 +687,12 @@ mod tests {
 
         #[test]
         fn on_timeout_not_exceed_max() {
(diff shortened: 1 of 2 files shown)


jacobkaufmann:
@carver this should address some of the issues raised in #69

PR #76, opened 2023-06-13
Prevent invalid ids being used with connect_with_cid/accept_with_cid

KolbyML:
for initiator if cid.recv.wrappingadd(1) != cid.send

It is an invalid cid configuration and will cause other clients to reset. Since this is invalid use of the library we should throw an error if someone tries to use an invalid cid configuration.

I also added the inverse case for ``accept_with_cid``

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -225,9 +225,16 @@ where
 
     pub async fn accept_with_cid(
         &self,
-        cid: ConnectionId<P>,
+        cid_recv: u16,
+        cid_peer: P,
         config: ConnectionConfig,
     ) -> io::Result<UtpStream<P>> {
+        let cid = ConnectionId {
+            send: cid_recv.wrapping_sub(1),
+            recv: cid_recv,
+            peer: cid_peer,
+        };
+
         let (stream_tx, stream_rx) = oneshot::channel();
         let accept = Accept {
             stream: stream_tx,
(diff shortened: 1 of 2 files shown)


jacobkaufmann:
good catch! ideally this should not even be possible. I see two obvious options. there may be others.

the first would alter the function parameters to the following:

```rust
pub async fn connect_with_cid(
        &self,
        cid_send: u16,
        cid_peer: P,
        config: ConnectionConfig,
    ) -> io::Result<UtpStream<P>> {
```

the second would modify [`ConnectionId`](https://github.com/ethereum/utp/blob/6c991ab5468e9e33610f317eaca380ca3c1766fa/src/cid.rs#L12) such that you cannot construct an instance with invalid values. that would involve making the relevant fields private (or preferably `pub(crate)`) and adding a `pub` constructor.

jacobkaufmann:
also worth a note that we should probably also enforce same for `accept_with_cid`. if we take the second option given above, then I think that would take care of that.

KolbyML:
Ok I implemented your suggested solution 1 since I thought it was cleaner. The PR is ready to review again :shipit:

src/socket.rs
```diff
@@ -269,9 +276,16 @@ where
 
     pub async fn connect_with_cid(
         &self,
-        cid: ConnectionId<P>,
+        cid_send: u16,
```
carver, line 279: I was noticing that different sides of the connection are used on connect and accept, and I was trying to understand the logic for why we would use one or the other. I looked at how we use it in trin, and the recv side is the natural one to use there. I'm sure you have a reason why `send` is the natural choice. I think picking one is likely to lead to confusion, either way, so I'm starting to think that the other solution Jacob suggested might be better.

The kind of mess of 100 and 101s in the test seem to align with that. It seems very easy to lose track of.

KolbyML: That would have similar problems, but I can take a shot at it, and will put it in a different PR

it would probably look like this

```
func new_connect(id: u16, is_sender: bool) {
    if (is_sender) {}
    else {}
}

So I think the same confusion would happen

KolbyML: Ok the alternative solution is in this PR https://github.com/ethereum/utp/pull/95

PR #77, opened 2023-06-14
fix: reorder arithmetic in SendBuffer::available to prevent overflow

KolbyML:
available overflows in ``send.rs`` since the add for the offset is done after (I verified this testing local and this fixing it).

superedes https://github.com/ethereum/utp/pull/66 since the review process never got resolved, it has been 2 weeks without a response from pr 66's author, and this overflow bug is an annoying distraction well testing other problems so it would be nice to get it merged.

### DIFF
--- a/src/send.rs
+++ b/src/send.rs
@@ -29,7 +29,7 @@ impl<const N: usize> SendBuffer<N> {
 
     /// Returns the number of bytes available in the buffer.
     pub fn available(&self) -> usize {
-        N - self.pending.iter().fold(0, |acc, x| acc + x.len()) + self.offset
+        N + self.offset - self.pending.iter().fold(0, |acc, x| acc + x.len())
     }
 
     /// Returns `true` if the buffer is empty.


jacobkaufmann:
have you been able to demonstrate that this overflows? #66 description reads "there is no need to panic when more values are pending than the buffer capacity", but due to the check [here](https://github.com/ethereum/utp/blob/6c991ab5468e9e33610f317eaca380ca3c1766fa/src/send.rs#L43), that situation should not be possible. in other words, the overflow should not occur because the total size of elements of `pending` should not exceed the buffer size `N`.

KolbyML:
If you want to reproduce it pull this branch https://github.com/KolbyML/utp/tree/failing-code

cd into the utp-client-interop folder

### Test ethereum/utp  send to bittorrent/libutp

Terminal 1
```
make rcrecv
```

Terminal 2
```
make rrsend
```

After you run this ``ethereum/utp`` outputs
```
1738Z DEBUG uTP{send=10826 recv=10825}: utp_rs::conn: uTP conn starting...
thread 'tokio-runtime-worker' panicked at 'attempt to subtract with overflow', /home/kolby/.cargo/registry/src/index.crates.io-6f17d22bba15001f/utp-rs-0.1.0-alpha.5/src/send.rs:32:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "RecvError(())" }', src/main.rs:62:47
```


After I reordered the calculation to ``N  + offset - pending`` instead of ``N  - pending + offset`` it stopped panicing

KolbyML:
You can replicate it with for i in 1..x I know x can be as low as 4k or 3k I was just using 30000 to test with

jacobkaufmann:
okay I see what's going on. nice!

PR #79, opened 2023-06-16
Added tracing::trace!() for "Write data"/"Received packet"

KolbyML:
I added debug traces for "Write data"/"Received packet". Because when I am debugging it is hard to tell if nodes are even sending anything or receiving anything.

These longs are simliar to what fluffy/utp and bittorrent/libutp

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -689,6 +689,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
     }
 
     fn on_packet(&mut self, packet: &Packet, now: Instant) {
+        tracing::trace!("Received {:?}", packet);
         let now_micros = crate::time::now_micros();
         self.peer_recv_window = packet.window_size();
 
@@ -1103,6 +1104,14 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
                 packet.payload().len() as u32,
             )
         };
+        tracing::trace!(
+            "Write data cid={} packetType={} pkSeqNr={} pkAckNr={} length={}",
+            packet.conn_id(),
+            packet.packet_type(),
+            packet.seq_num(),
+            packet.ack_num(),
+            packet.encoded_len()
+        );
 
         sent_packets.on_transmit(packet.seq_num(), packet.packet_type(), payload, len, now);
         unacked.insert_at(packet.seq_num(), packet.clone(), sent_packets.timeout());
--- a/src/packet.rs
+++ b/src/packet.rs
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::fmt::{Debug, Formatter};
 
 /// Size of an encoded uTP header in bytes.
 const PACKET_HEADER_LEN: usize = 20;
@@ -129,6 +130,20 @@ pub enum PacketType {
     State,
     Reset,
     Syn,
+}
+
+impl fmt::Display for PacketType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            Self::Data => "ST_DATA".to_string(),
+            Self::Fin => "ST_FIN".to_string(),
+            Self::State => "ST_STATE".to_string(),
+            Self::Reset => "ST_RESET".to_string(),
+            Self::Syn => "ST_SYN".to_string(),
+        };
+
+        write!(f, "{s}")
+    }
 }
 
 impl TryFrom<u8> for PacketType {
@@ -404,7 +419,7 @@ impl SelectiveAck {
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct Packet {
     header: PacketHeader,
     selective_ack: Option<SelectiveAck>,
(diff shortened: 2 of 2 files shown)


src/conn.rs
```diff
@@ -689,6 +689,15 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
     }
 
     fn on_packet(&mut self, packet: &Packet, now: Instant) {
+        tracing::debug!("Received packet cid={} packetType={} seqNr={} ackNr={} timestamp={} timestampDiff={} remoteWindow={}",
```
carver, line 692: Did you try using the debug printout `{:?}` of the packet? Seems cleaner than manually extracting each specific field.

KolbyML: Because it doesn't look at clean and has not clean
```
Received packet cid=201 packetType=ST_DATA seqNr=25874 ackNr=6553 timestamp=1612960711 timestampDiff=23325 remoteWindow=1048576
```

^ this is what I have now

If I change it to what you are wondering
``tracing::debug!("Received packet {:?}", packet,);``

```
 Received packet Packet { header: PacketHeader { packet_type: Data, version: One, extension: None, conn_id: 201, ts_micros: 1727574141, ts_diff_micros: 33980, window_size: 1048576, seq_num: 61519, ack_num: 20696 }, selective_ack: None, payload: [254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254] }
```
 
 or
 
 ```
 Received packet Packet { header: PacketHeader { packet_type: State, version: One, extension: None, conn_id: 200, ts_micros: 1727643848, ts_diff_micros: 69684, window_size: 1045581, seq_num: 20697, ack_num: 61517 }, selective_ack: None, payload: [] }
 ```
 
 So using ``{:?}`` makes the debug log a lot less readable and also just has a bunch of not needed text like ``Packet { header: PacketHeader {``

KolbyML: Also the data log using ``{:?}`` takes up my entire terminal which is unidle state takes up 2 lines

Where mine shown above takes only 1 line

carver: So I think the preferred approach here would be to add a custom Debug formatter with these fields, and use it in both directions. Then, it would be useful anywhere that we want to log out a packet's info.

Also, `trace` seems like the appropriate level for a log that prints out every single packet.

KolbyML: It is using trace I just cropped that out in my examples since I thought they were distracting for an example context. 

Thank you for the suggestion for the custom debug trait I didn't even think of it thank you 👍

KolbyML: Ok I made that adjustment for ``on_packet``, I think the debug in ``on_transmit`` should stay the same though

carver (dismissed):
Cool, yeah the display formatters look good to me.

My last bit is that I think both of these should be `trace` level. Printing every packet will be *very* noisy in many scenarios, and debug logs probably shouldn't spew quite that much info.

That's my last blocker for merging. 🎉

PR #81, opened 2023-06-16
test: run 2k concurrent transfers with 1MB of data

carver:
This used to be a PR that grouped up all the changes needed to pass the high-concurrency tests. Now, with all the changes merged, the test is the only commit left, and it passes against master.

---
Previous writeup, for posterity

## Handle High Concurrency

The minimal functional changes required to pass the high-volume socket test, plus some logs, pulled from #69 

The commits should be fairly clean to inspect individually, if you're into that kind of thing.

The functional changes are:
- Bias the connection event loop, to prefer handling inbound packets. Previously, it was possible to randomly choose to timeout a connection, even if an inbound packet was waiting on the queue.
- Use exponential backoff when sending SYN times out (the test tried a *bunch* of SYNs at once, exponential backoff helps recover from this stressed condition)
- A bit more wiggle room in the default config (longer idle timeout, more connection retries, more DATA packet retries)
- Bias the socket event loop, to prefer handling socket reads (usually UDP). UDP packet drops were happening a lot on high-volume tests. It is much more work to recover from drops, than from delays to outbound sending (a side effect of preferring reads), and uTP already nicely handles delays.
- Finally, add a limit to the number of connections. This is applied on new outbound connections, but considers the total connection count.

Altogether, these changes reliably cause the new test to pass, even when bumped up to 10,000 simultaneously-launched transfers. (Though the simultaneous bit is mostly a red-herring now, since the stream prevents them from actually starting to run at the same time)

There was a probabilistic success improvement for most of these changes, though the connection limiting was the final piece that gave apparent reliability. It might be tempting to drop some of these changes and just stick to the connection limit. However, since the connection limit is a configurable value, we want to support high connection rates as much as possible. So it seems worth keeping the earlier functional changes, to do a better job if users try high values.

Notably *not* present is the "stress tracking" or "busy/idle tracking" from the socket. Both provided interesting information, but were not reliable enough to use to apply back-pressure. The added code complexity wasn't worth the (admittedly quite valuable) logs, IMO.

The logging changes are mostly not visible except in stress/failure situations, but they are very helpful for debugging, so I think we should keep them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,5 +22,8 @@ tracing = { version = "0.1.37", features = ["std", "attributes", "log"] }
 
 [dev-dependencies]
 quickcheck = "1.0.3"
-tracing-subscriber = "0.3.16"
 tokio = { version = "1.25.0", features = ["test-util"] }
+tracing-subscriber = "0.3.16"
+
+[profile.test]
+opt-level = 3
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -1,105 +1,96 @@
+use futures::stream::{FuturesUnordered, StreamExt};
 use std::net::SocketAddr;
 use std::sync::Arc;
+
+use tokio::task::JoinHandle;
+use tokio::time::Instant;
 
 use utp_rs::cid;
 use utp_rs::conn::ConnectionConfig;
 use utp_rs::socket::UtpSocket;
 
-#[tokio::test(flavor = "multi_thread")]
+const TEST_DATA: &[u8] = &[0xf0; 1_000_000];
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 16)]
 async fn socket() {
     tracing_subscriber::fmt::init();
 
-    let conn_config = ConnectionConfig::default();
-
-    let data_one = vec![0xef; 8192 * 2 * 2];
-    let data_one_recv = data_one.clone();
+    tracing::info!("starting socket test");
 
     let recv_addr = SocketAddr::from(([127, 0, 0, 1], 3400));
+    let send_addr = SocketAddr::from(([127, 0, 0, 1], 3401));
+
     let recv = UtpSocket::bind(recv_addr).await.unwrap();
     let recv = Arc::new(recv);
-
-    let send_addr = SocketAddr::from(([127, 0, 0, 1], 3401));
     let send = UtpSocket::bind(send_addr).await.unwrap();
     let send = Arc::new(send);
+    let mut handles = FuturesUnordered::new();
 
-    let recv_one_cid = cid::ConnectionId {
-        send: 100,
-        recv: 101,
+    let start = Instant::now();
+    let num_transfers = 2000;
+    for i in 0..num_transfers {
+        // step up cid by two to avoid collisions
+        let handle =
+            initiate_transfer(i * 2, recv_addr, recv.clone(), send_addr, send.clone()).await;
+        handles.push(handle.0);
+        handles.push(handle.1);
+    }
+
+    while let Some(res) = handles.next().await {
+        res.unwrap();
+    }
+    let elapsed = Instant::now() - start;
+    let megabits_sent = num_transfers as f64 * TEST_DATA.len() as f64 * 8.0 / 1_000_000.0;
+    let transfer_rate = megabits_sent / elapsed.as_secs_f64();
+    tracing::info!("finished real udp load test of {} simultaneous transfers, in {:?}, at a rate of {:.0} Mbps", num_transfers, elapsed, transfer_rate);
+}
+
+async fn initiate_transfer(
+    i: u16,
+    recv_addr: SocketAddr,
+    recv: Arc<UtpSocket<SocketAddr>>,
+    send_addr: SocketAddr,
+    send: Arc<UtpSocket<SocketAddr>>,
+) -> (JoinHandle<()>, JoinHandle<()>) {
+    let conn_config = ConnectionConfig::default();
+    let initiator_cid = 100 + i;
+    let responder_cid = 100 + i + 1;
+    let recv_cid = cid::ConnectionId {
+        send: initiator_cid,
+        recv: responder_cid,
         peer: send_addr,
     };
-    let send_one_cid = cid::ConnectionId {
-        send: 101,
-        recv: 100,
+    let send_cid = cid::ConnectionId {
+        send: responder_cid,
+        recv: initiator_cid,
         peer: recv_addr,
     };
 
-    let recv_arc = Arc::clone(&recv);
-    let recv_one_handle = tokio::spawn(async move {
-        let mut stream = recv_arc
-            .accept_with_cid(recv_one_cid, conn_config)
-            .await
-            .unwrap();
+    let recv_handle = tokio::spawn(async move {
+        let mut stream = recv.accept_with_cid(recv_cid, conn_config).await.unwrap();
         let mut buf = vec![];
-        let n = stream.read_to_eof(&mut buf).await.unwrap();
-        tracing::info!(cid.send = %recv_one_cid.send, cid.recv = %recv_one_cid.recv, "read {n} bytes from uTP stream");
+        let n = match stream.read_to_eof(&mut buf).await {
+            Ok(num_bytes) => num_bytes,
+            Err(err) => {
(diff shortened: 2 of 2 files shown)


src/socket.rs
```diff
@@ -242,14 +253,36 @@ where
         }
     }
 
+    async fn add_outbound_connection(&self, cid: &ConnectionId<P>, events_tx: mpsc::UnboundedSender<StreamEvent>) {
+        loop {
+            let num_connections = self.conns.read().unwrap().len();
+            if num_connections < self.max_conns {
+                break;
+            }
+            self.outbound_dropped.notified().await;
```
carver, line 262: Is this holding the read() lock across the await? Do we need something like:

```suggestion
            let num_connections = {
                self.conns.read().unwrap().len()
            };
            if num_connections < self.max_conns {
                break;
            }
            self.outbound_dropped.notified().await;
```

It hasn't seemed to cause problems during testing, but it's obviously worth double-checking.

src/socket.rs
```diff
@@ -27,12 +28,16 @@ pub struct UtpSocket<P> {
...
     socket_events: mpsc::UnboundedSender<SocketEvent<P>>,
+    /// Maximum allowed connections, which only restricts outbound connections, but counts all
+    /// active connections.
+    max_conns: usize,
+    outbound_dropped: Arc<Notify>,
 }
 
 impl UtpSocket<SocketAddr> {
-    pub async fn bind(addr: SocketAddr) -> io::Result<Self> {
+    pub async fn bind(addr: SocketAddr, max_conns: usize) -> io::Result<Self> {
```
carver, line 38: Add a docstring with some hints about how to choose `max_conns`. Roughly:
- 100 seems like a reasonable starting place for a lot of use cases, 1k is probably a bit too high, depending on your setup, 10k will likely cause overload, 1-5 is probably so small that you won't be getting the full potential of the library
- A too-high value can cause connection timeouts, and various logs about packets taking too long to be processed locally
- A too-low value will cause connections to delay for a long time, while waiting for `connect()` or `connect_with_cid()` to return. You will not experience the full performance of the library

src/conn.rs
```diff
@@ -101,14 +101,14 @@ pub struct ConnectionConfig {
...
+        let max_idle_timeout = Duration::from_secs(30);
         Self {
-            max_conn_attempts: 3,
+            max_conn_attempts: 5,
             max_idle_timeout,
             max_packet_size: congestion::DEFAULT_MAX_PACKET_SIZE_BYTES as u16,
             initial_timeout: congestion::DEFAULT_INITIAL_TIMEOUT,
             min_timeout: congestion::DEFAULT_MIN_TIMEOUT,
-            max_timeout: max_idle_timeout,
+            max_timeout: max_idle_timeout / 4,
```
carver, line 111: Move the 4 to a file constant, and explain the reasoning

src/socket.rs
```diff
@@ -27,12 +28,16 @@ pub struct UtpSocket<P> {
     cid_gen: Mutex<StdConnectionIdGenerator<P>>,
     accepts: mpsc::UnboundedSender<(Accept<P>, Option<ConnectionId<P>>)>,
     socket_events: mpsc::UnboundedSender<SocketEvent<P>>,
+    /// Maximum allowed connections, which only restricts outbound connections, but counts all
+    /// active connections.
+    max_conns: usize,
```
jacobkaufmann, line 33: what is the rationale on restricting only outbound connections?

jacobkaufmann (commented):
nice work! left some comments on my first pass.

it might be easier to break these up into smaller PRs, since it will become difficult to keep the changes for PR back-and-forth organized, and I'd prefer to squash all commits regardless

I would see a minimum of 3 sets of changes:

* concurrent connection limits
* `tokio::select` biasing
* exponential backoff for outgoing connection establishment

src/conn.rs
```diff
@@ -249,9 +249,16 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         tokio::pin!(idle_timeout);
         loop {
             tokio::select! {
+                biased;
```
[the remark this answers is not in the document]
carver: Ah, that's an interesting approach that hadn't occurred to me.

The tradeoffs I'm seeing on that choice are:

Pros of switching to `try_recv`
- Fewer opportunities to mess up the branch order in subtle ways
- `biased` is a less-familiar approach, it requires more understanding from the contributors

Cons of switching to `try_recv`
- Slight performance loss (`biased` is a bit faster)
- In maintenance, we'd want to make sure that any path that resets `idle_timeout` is also double-checked in the timeout branch. For example here, I guess we'd also want to check the `writes.try_recv()`
- More lines of code: the timeout processing gets messier. Do we just restart the loop if there is a packet waiting on the queue (in which case we might loop through a random number of times before hitting the branch we want), or we could actually process the packet in the idle-timeout branch, which seems surprising and leaves room open for bugs where someone makes a change in the packet-processing branch that doesn't get copied to the timeout branch.

Based on these tradeoffs, I'm inclined to keep the `biased` approach.

> do we need to move any of the other branches (e.g. shutdown)?

I haven't found any other important races, when both branches are notified, and we need to process one before the other. We might get some performance benefit of processing writes before reads, or the other way around, but I can't see any place where it would cause a problem. Eventually, one branch will always drain out and we'll move onto the next one. It doesn't seem we'll get stuck in any particular branch.

So I generally favored changing the order as little as possible. What's your thinking on moving the shutdown branch?

src/conn.rs
```diff
@@ -285,10 +286,31 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
                 _ = self.writable.notified() => {
                     self.process_writes(Instant::now());
                 }
+                Some(Ok(timeout)) = self.unacked.next() => {
+                    let (_seq, packet) = timeout;
+                    //tracing::debug!(seq, ack = %packet.ack_num(), packet = ?packet.packet_type(), "timeout");
```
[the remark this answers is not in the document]
carver: Ah yes, it was just very noisy during testing. I'll add it back.

src/conn.rs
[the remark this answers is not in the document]
carver: Another potential benefit of keeping `biased` is using it to do something like: process all incoming packets before sending the SYN that ACKs the very latest one.

src/conn.rs
```diff
@@ -1106,9 +1133,10 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
 
         sent_packets.on_transmit(packet.seq_num(), packet.packet_type(), payload, len, now);
         unacked.insert_at(packet.seq_num(), packet.clone(), sent_packets.timeout());
-        socket_events
-            .send(SocketEvent::Outgoing((packet, dest.clone())))
-            .expect("outgoing channel should be open if connection is not closed");
+        let send_result = socket_events.send(SocketEvent::Outgoing((packet, dest.clone())));
+        if send_result.is_err() {
+            tracing::error!("Cannot transmit packet: socket_events closed");
+        }
```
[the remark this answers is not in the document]
carver: 👍🏻 to unifying the log level. It's fairly obnoxious to have a passing test end with an error log, and it happens pretty regularly, so I'd like to make them both warn.

src/socket.rs
carver: Hm, this one is a little fuzzy. I suppose that:
- outgoing connections feel delayable until we want to send the data, but inbound connections were presumably initiated by someone else, so we can't just wait as long as we want to accept
- we might have a preference to receive data (help selves) over sending it (help others), if we are resource constrained

Any thoughts on this? I suppose the alternative would be to add the same pausing mechanism into the accepts.

jacobkaufmann: I think the first point is more important than the second.

if we delay accepting a connection, then the initiating endpoint may give up on the connection before we accept, and then when we accept the connection we will consume a connection until the idle timeout.

uTP streams are bidirectional, and the accepting endpoint of a uTP stream is not necessarily the endpoint receiving data, so I don't think we can make assumptions about one endpoint being the sender and the other being the receiver.

that said, I think it's okay to only limit outgoing connections if we make it clear from the config that it is not total connections. if it becomes evident that an unlimited number of incoming (i.e. accepted) connections are harming outgoing connections, then we can address it with a different approach.

carver:
Test error discussed in #103

PR #88, opened 2023-06-21
Close without error, when only FIN-ACK is missing on timeout, and related logs

carver:
Split from #81 
- Catch a special case that doesn't need to be a warning: Dual FINs, all data ACKd
  - set the Close status err to None
  - log it at debug level only, it's not worth spamming the logs for
- Add a Debug formatter to the connection State
- Add state info to the idle timeout log

Optional extra:
- [x] validate that the stream closes without error if only fin-ack is missing (basic structure: wait for read to complete, make the sender's outgoing packets drop, call shutdown from the recipient. The sender should close with no error, the recipient should close with an error)

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -328,13 +328,16 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
                             State::Closing { local_fin, remote_fin, .. } => unacked.len() == 1 && local_fin.is_some() && &local_fin.unwrap() == unacked.last().unwrap() && remote_fin.is_some(),
                             _ => false,
                         };
-                        if finished {
+
+                        let err = if finished {
                             tracing::debug!(?self.state, ?unacked, "idle timeout expired, but only missing ACK for local FIN");
+                            None
                         } else {
                             tracing::warn!(?self.state, ?unacked, "closing, idle for too long...");
-                        }
+                            Some(Error::TimedOut)
+                        };
 
-                        self.state = State::Closed { err: Some(Error::TimedOut) };
+                        self.state = State::Closed { err };
                     }
                 }
                 _ = &mut shutdown, if !shutting_down => {
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -317,8 +317,22 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
                 }
                 () = &mut idle_timeout => {
                     if !std::matches!(self.state, State::Closed { .. }) {
+                        // Warn that we are quitting the connection, due to a lack of activity.
+
+                        // If both endpoints have exchanged FINs, and our FIN is the only
+                        // unacknowledged packet, then do not emit a warning. It's not a big deal
+                        // that their STATE for our FIN got dropped: we handled their FIN anyway.
+
                         let unacked: Vec<u16> = self.unacked.keys().copied().collect();
-                        tracing::warn!(?unacked, "idle timeout expired, closing...");
+                        let finished = match self.state {
+                            State::Closing { local_fin, remote_fin, .. } => unacked.len() == 1 && local_fin.is_some() && &local_fin.unwrap() == unacked.last().unwrap() && remote_fin.is_some(),
+                            _ => false,
+                        };
+                        if finished {
+                            tracing::debug!(?self.state, ?unacked, "idle timeout expired, but only missing ACK for local FIN");
+                        } else {
+                            tracing::warn!(?self.state, ?unacked, "closing, idle for too long...");
+                        }
 
                         self.state = State::Closed { err: Some(Error::TimedOut) };
                     }
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -92,6 +92,25 @@ enum State<const N: usize> {
     },
 }
 
+impl<const N: usize> fmt::Debug for State<N> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Connecting(_) => write!(f, "State::Connecting"),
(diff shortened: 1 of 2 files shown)


src/conn.rs
```diff
@@ -287,8 +306,24 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
...
-                        tracing::warn!(?unacked, "idle timeout expired, closing...");
+                        if let State::Closing { local_fin, remote_fin, .. } = self.state {
+                            if unacked.len() == 1 && local_fin.is_some() && &local_fin.unwrap() == unacked.last().unwrap() && remote_fin.is_some() {
+                                tracing::debug!(?self.state, ?unacked, "Idle timeout with both FINs sent, with only FIN unacked.");
+                            } else {
+                                tracing::warn!(?self.state, ?unacked, "quitting idle connection...");
+                            }
+                        } else {
+                            tracing::warn!(?self.state, ?unacked, "quitting idle connection...");
+                        }
```
carver, line 326: I don't like this duplicate warning line. Does something like this seem better?
```rs
let finished = match self.state {
  Closing {} => unacked.len() == 1 && <etc>
  _ => false
};
if finished {
  tracing::debug!(...)
} else {
  tracing::warn!(...)
}

File: src/conn.rs [hunk 2]
```diff
@@ -287,8 +306,24 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
...
                         let unacked: Vec<u16> = self.unacked.keys().copied().collect();
-                        tracing::warn!(?unacked, "idle timeout expired, closing...");
+                        let finished = match self.state {
+                            State::Closing { local_fin, remote_fin, .. } => unacked.len() == 1 && local_fin.is_some() && &local_fin.unwrap() == unacked.last().unwrap() && remote_fin.is_some(),
+                            _ => false,
+                        };
+                        if finished {
+                            tracing::debug!(?self.state, ?unacked, "Idle timeout with both FINs sent, with only FIN unacked.");
+                        } else {
+                            tracing::warn!(?self.state, ?unacked, "quitting idle connection...");
```
jacobkaufmann, line 325: nit:

```suggestion
                            tracing::warn!(?self.state, ?unacked, "closing idle connection...");
```

src/conn.rs
```diff
@@ -287,8 +306,24 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
...
+                        // There is no more work to do.
+
                         let unacked: Vec<u16> = self.unacked.keys().copied().collect();
-                        tracing::warn!(?unacked, "idle timeout expired, closing...");
+                        let finished = match self.state {
+                            State::Closing { local_fin, remote_fin, .. } => unacked.len() == 1 && local_fin.is_some() && &local_fin.unwrap() == unacked.last().unwrap() && remote_fin.is_some(),
+                            _ => false,
+                        };
+                        if finished {
+                            tracing::debug!(?self.state, ?unacked, "Idle timeout with both FINs sent, with only FIN unacked.");
```
jacobkaufmann, line 323: prefer lowercase and no terminal punctuation

```suggestion
                            tracing::debug!(?self.state, ?unacked, "idle timeout expired, but only missing ACK for local FIN");
```

src/conn.rs
```diff
@@ -287,8 +306,24 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
                 }
                 () = &mut idle_timeout => {
                     if !std::matches!(self.state, State::Closed { .. }) {
+                        // Warn that we are giving up on the connection due to a lack of activity.
+
+                        // First, we look for this special situation:
+                        // Our outgoing FIN was sent, but the incoming ACK was dropped.
+                        // This timeout is not worthy of a warning log, because everything
+                        // basically went fine: all data was sent and ACK'd.
+                        // There is no more work to do.
```
jacobkaufmann, line 315: I think this is a bit more clear/concise

```suggestion
                        // If both endpoints have exchanged FINs, and our FIN is our only unacknowledged
                        // packet, then do not emit a warning, since a single STATE packet could have simply
                        // been dropped
```

src/conn.rs
carver: So the reason for "quit" here is to convey that the connection ended abnormally. Using "close" hints that it was a normal exit.

This isn't a hill I'm going to die on, but I do like quit a bit better. What are the implications of "close" and "quit" for you?

src/conn.rs
carver: Ok, I took another pass.

src/conn.rs
jacobkaufmann: well all connections "close". that is the terminal state. to me, the bit about "idle connection" conveys the additional context.

src/conn.rs
```diff
@@ -1378,4 +1411,51 @@ mod test {
         conn.on_reset();
         assert!(std::matches!(conn.state, State::Closed { err: None }));
     }
+
+    #[test]
+    fn state_debug_format() {
```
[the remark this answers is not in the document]
carver: I think the upside of testing even non-critical features like this is that it:
- helps junior contributors catch their own mistakes if a refactor messes something up (even if it's not-critical, I don't want it to change by accident)
- specifies the custom output, and helps reviewers identify any intentional changes to that "spec"
- sets a standard that new things in the code base should be tested in CI

The costs are:
- increased test time (measured in single milliseconds, at most)
- maintenance cost for unrelated changes (like adding a new field into a status... although it's certainly a moment we would want to consider whether to include the new field in the debug output. But there are surely other examples of unrelated code changes that would require an update to the test)

The costs are so infrequent and low-effort, that I feel the benefits outweigh them.

carver:
Ok, addressed all the feedback, and added a test showing that close is returned without error in the specific case. Merging!

Issue #89, opened 2023-06-22
Change shutdown() to block until the write data is sent & acked


carver:
Callers (like `trin`) are interested in whether the written data has been sent (and ACK'd by the receiver). One option would be to expose a new `flush()` method that waits until the buffer is empty. It adds new complexity in a write/flush/write scenario that is not of interest (to `trin` at least), so let's just fold it into 'shutdown()' and not add a new API.

Instead, the new behavior would be that `shutdown()` will wait to return until all previously-written data has been sent and acknowledged (or, of course, will return early if the connection exits due to a failure). Currently, shutdown returns optimistically, then the connection tries to wrap up in the background.

One disadvantage of this approach is that the shutdown might report an error while handling the "wrap-up" steps, or if the peer disconnected between flushing all data and our attempt to shutdown. But that ought to be called a successful flush, from `trin`s point of view. So we might have to carefully parse the different kinds of error coming out of a shutdown (or add new types of error). Using a new `flush()` method instead would mean we could return successfully and immediately if our peer had previously ACKd all data, even if they already hung up. That would be convenient and reliable for trin.

PR #91, opened 2023-06-24
Log error, don't crash, if socket closes mid-send

carver:
Split from #81 

Error will always be the same type, no need to log it.

Other conversation: https://github.com/ethereum/utp/pull/81#discussion_r1234472553

Ideally, we would detect somehow that it was a normal shutdown, and we would just silently exit. But if it happens at a surprising time, then it is valuable to have the log. Warn seems like about the right level for that, especially since it typically doesn't trigger very often.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -746,9 +746,11 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         match packet.packet_type() {
             PacketType::Syn | PacketType::Fin | PacketType::Data => {
                 if let Some(state) = self.state_packet() {
-                    self.socket_events
-                        .send(SocketEvent::Outgoing((state, self.cid.peer.clone())))
-                        .expect("outgoing channel should be open if connection is not closed");
+                    let event = SocketEvent::Outgoing((state, self.cid.peer.clone()));
+                    if self.socket_events.send(event).is_err() {
+                        tracing::warn!("Cannot transmit state packet: socket closed channel");
+                        return;
+                    }
                 }
             }
             PacketType::State | PacketType::Reset => {}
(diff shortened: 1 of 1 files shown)


jacobkaufmann (commented):
maybe we don't want to panic, but I don't think a log is sufficient. if the channel to the socket is closed, then the connection cannot progress, since it will be unable to communicate data and/or acknowledgements. if we don't panic, then we should close with an error.

carver:
I see where you're going. I think what's happening is that there's a bug that causes the socket to disappear before the connection completes. (This is possibly something that could be fixed by adding an internal flush, like #89 )

But the current behavior of panicking is unacceptable, because it causes flaky test failures. Changing the behavior to cause an error in the shutdown or connection result would have the same problem, since the test unwrap's these results (correctly, IMO).

I am starting to run out of time to work on utp stuff, other things need my attention. How about I add an issue to track down why the connection sometimes keeps running after the socket is dropped, and we merge this patch until then?

carver:
Ok, I tried something that closes the connection with an error if the socket is unavailable for writes. I kind of don't think it's worth this much extra error handling in order to shut down the connection, just to deal with an occasional flaky issue that's only showing up during tests (some kind of race with shutting the connection down).

My intuition is to just go back to the first commit in this PR, but I wanted to show you this in case you had a better idea how to implement shutting down the connection.

carver:
Ok, my next thought is that we can go *back* to panicking everywhere, except while the connection is handling the final steps of closing down (which I think is where it's happening anyway). It still means having `transmit()` return a result, but we can just `expect()` it everywhere except for where we handle setting the state to `Closed`.

Issue #38, opened 2023-03-18
add socket-level config


jacobkaufmann:
add socket-level configuration to constructor(s)

the config could include the following:

* max UDP packet size
* max connections

carver:
I don't think we're ever practically changing connection configurations from one connection to another. I'm starting to think maybe everything could be moved to a Socket level config. That seems easier for the caller than managing two separate configs. What do you think?

carver:
Splitting the Config into two parts seems to push us in the unpleasant direction of #92.

I'm continuing to lean toward fully removing ConnectionConfig and replacing it with a SocketConfig (or StreamConfig?)

PR #92, opened 2023-06-26
Peel off a ConnectionConfig value into SocketConfig

carver:
This split is hairy and seems error-prone. This is one reason I'm heavily considering just moving the whole ConnectionConfig into a SocketConfig. See https://github.com/ethereum/utp/issues/38


carver:
Closing. If we're going to move a config value at all, I expect it to be in the direction of #93 instead.

PR #94, opened 2023-06-27
Add API to get the current number of connections

carver:
Useful for callers to get statistics on how many active connections are in flight. This is especially true since they need some awareness of the maximum number of connections in the configuration.

### DIFF
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -101,3 +101,93 @@ async fn socket() {
     one.unwrap();
     two.unwrap();
 }
+
+// Test that a new socket has zero connections
+#[tokio::test]
+async fn test_empty_socket_conn_count() {
+    let socket_addr = SocketAddr::from(([127, 0, 0, 1], 3402));
+    let socket = UtpSocket::bind(socket_addr).await.unwrap();
+    assert_eq!(socket.num_connections(), 0);
+}
+
+// Test that a socket returns 2 from num_connections after connecting twice
+#[tokio::test]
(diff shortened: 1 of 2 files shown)


tests/socket.rs
```diff
@@ -101,3 +101,93 @@ async fn socket() {
...
+#[tokio::test]
+async fn test_empty_socket_conn_count() {
+    let socket_addr = SocketAddr::from(([127, 0, 0, 1], 3402));
+    let socket = UtpSocket::bind(socket_addr).await.unwrap();
+    assert_eq!(socket.num_connections(), 0);
+}
+
+// Test that a socket returns 2 from num_connections after connecting twice
+#[tokio::test]
+async fn test_socket_reports_one_connection() {
```
jacobkaufmann, line 115: IMO we should remove these tests. I don't see any benefit for the nearly extra 100 LOC

carver: Hm, not sure where to go with this. I agree that the tests feel a bit long. I don't agree that having zero test coverage of the API is preferable. So unless we have a third option, having verbose tests seems is the clear winner to me.

I've never really seen a project that I trust that prefers skipping tests altogether. Maybe we can build some testing tools to make it easier and more succinct to write the tests? Having at least basic test coverage of public APIs is a core requirement to move toward stability, from my perspective.

carver:
We talked offline. I'm going to be a bit more liberal with merges, going forward. I'll still try to pay special attention to not over-fitting utp to trin's needs.

PR #101, opened 2023-08-07
Allow for mutable state in Socket send/recv

carver:
Use `&mut self` in `send_to` and `recv_from` methods.

This allows Socket implementations to manage their own local state when handling send and receive. An example use case of this is implementing a socket that uses a local channel. Send and receive need to access the channel mutably.

This will also used by trin, as detailed & linked in: https://github.com/ethereum/utp/pull/98#discussion_r1283504457


src/udp.rs
```diff
@@ -10,18 +10,20 @@ use crate::cid::ConnectionPeer;
...
-    async fn recv_from(&self, buf: &mut [u8]) -> io::Result<(usize, P)>;
+    async fn recv_from(&mut self, buf: &mut [u8]) -> io::Result<(usize, P)>;
 }
 
 #[async_trait]
 impl AsyncUdpSocket<SocketAddr> for UdpSocket {
-    async fn send_to(&self, buf: &[u8], target: &SocketAddr) -> io::Result<usize> {
-        self.send_to(buf, *target).await
+    async fn send_to(&mut self, buf: &[u8], target: &SocketAddr) -> io::Result<usize> {
+        // We need a different name than UdpSocket's send_to, or else we get a stack overflow.
```
jacobkaufmann, line 22: can we remove this line?

carver:
Anything else?

jacobkaufmann (approved):

PR #107, opened 2023-08-24
Acknowledge sent packets when receiving a FIN

### LINKED ISSUE #103: Seeing close() error when peer drops quickly after read_to_eof()
When testing with high concurrency, like in #81 -- I often see the sender fail, with a warning log:

> 2023-08-09T22:27:24.490262Z  WARN uTP{send=975 recv=974}: utp_rs::conn: closing, idle for too long... self.state=State::Closing { local_fin: Some(41692), remote_fin: Some(52419) } unacked=[41691, 41683, 41662, 41663, 41689, 41690, 41667, 41674, 41692, 41668, 41677, 41673, 41672, 41665, 41666, 41664, 41685, 41687, 41670, 41686, 41682, 41681, 41676, 41680, 41688, 41679, 41684, 41678, 41669, 41675, 41671]

Notice that earlier, the data recipient posted:
> 2023-08-09T22:26:54.453966Z  INFO socket: read 1000000 bytes from uTP stream cid.send=974 cid.recv=975

So from the recipient's point of view, everything went fine and exited smoothly. Also note that both FINs have been exchanged.

**Edit: The ack number in the inbound FIN is being ignored.**

---
Old irrelevant thoughts:

This opens a philosophical question about when we can consider the connection successful. We already did this little nudge in the direction of requiring the FIN-ACK to be a successful connection, but there may be a larger shift that makes sense.

From the utp spec:
> Finalize the connection. This is the last packet. It closes the connection, similar to TCP FIN flag. This connection will never have a sequence number greater than the sequence number in this packet. The socket records this sequence number as eof_pkt. This lets the socket wait for packets that might still be missing and arrive out of order even after receiving the ST_FIN packet.

So my latest take is that: from a sender's point of view, if we receive a FIN (and there are no missing sequence numbers before the FIN we received), then we can immediately drop all records of unacked sent packets. The connected peer has already decided to not receive anything else from us anyway.

Thoughts?

carver:
Fix #103

Two changes:
- refactor to make the next change easier (remove an unnecessary Option)
- **Main Change**: acknowledge sent packets using `ack_num` in FIN and DATA packets

More details on main change:

The ack_nr on the received packet was previously ignored. In some scenarios, like with dropped packets or different utp implementations, that would cause the sender to wait indefinitely for the recipient to ACK state. But the recipient had already acked with the FIN (been ignored), and disconnected. Then the sender was waiting the full connection timeout to realize it.

Even worse, if more than just the final outbound FIN was acknowledged in the inbound FIN, the sender was considering the data transfer incomplete. This was causing errors in #81 .

With this commit:
- When the recipient acks the final FIN with a FIN, the sender doesn't get stuck waiting for the timeout, and returns from close() immediately.
- When the recipient acks some data with the FIN, the sender now returns from close() with an OK instead of Err.

---

I tested this on a #81 locally, and (only) with this patch, I was able to get 10 successful test runs in a row, even when doubling the concurrency to 2000 connections.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -14,8 +14,7 @@ use crate::event::{SocketEvent, StreamEvent};
 use crate::packet::{Packet, PacketBuilder, PacketType, SelectiveAck};
 use crate::recv::ReceiveBuffer;
 use crate::send::SendBuffer;
-use crate::sent::SentPackets;
-use crate::seq::CircularRangeInclusive;
+use crate::sent::{SentPackets, SentPacketsError};
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum Error {
@@ -762,22 +761,22 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
 
         match packet.packet_type() {
             PacketType::Syn => self.on_syn(packet.seq_num()),
-            PacketType::State => {
-                let delay = Duration::from_micros(packet.ts_diff_micros().into());
-                self.on_state(
-                    packet.seq_num(),
-                    packet.ack_num(),
-                    packet.selective_ack(),
-                    delay,
-                    now,
-                );
-            }
+            PacketType::State => self.on_state(packet.seq_num(), packet.ack_num()),
             PacketType::Data => self.on_data(packet.seq_num(), packet.payload()),
-            PacketType::Fin => {
-                self.on_fin(packet.seq_num(), packet.payload());
-            }
-            PacketType::Reset => {
-                self.on_reset();
+            PacketType::Fin => self.on_fin(packet.seq_num(), packet.payload()),
+            PacketType::Reset => self.on_reset(),
+        }
+
+        // Mark sent packets as acked, using the received ack_num
+        match packet.packet_type() {
+            PacketType::Syn | PacketType::Reset => {}
+            PacketType::State | PacketType::Data | PacketType::Fin => {
+                let delay = Duration::from_micros(packet.ts_diff_micros().into());
+                if let Err(err) =
+                    self.process_ack(packet.ack_num(), packet.selective_ack(), delay, now)
+                {
+                    tracing::warn!(%err, ?packet, "ack does not correspond to known seq_num");
+                }
             }
         }
 
@@ -830,6 +829,36 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         }
     }
 
+    fn process_ack(
+        &mut self,
+        ack_num: u16,
+        selective_ack: Option<&SelectiveAck>,
+        delay: Duration,
+        now: Instant,
+    ) -> Result<(), Error> {
+        match &mut self.state {
+            State::Established { sent_packets, .. } | State::Closing { sent_packets, .. } => {
+                match sent_packets.on_ack(ack_num, selective_ack, delay, now) {
+                    Ok((full_acked, selected_acks)) => {
+                        self.unacked.retain(|seq, _| !full_acked.contains(*seq));
+                        for seq_num in selected_acks {
+                            self.unacked.remove(&seq_num);
+                        }
+                        Ok(())
+                    }
+                    Err(err) => match err {
+                        SentPacketsError::InvalidAckNum => {
+                            let err = Error::InvalidAckNum;
+                            self.reset(err);
+                            Err(err)
+                        }
+                    },
+                }
+            }
+            State::Connecting(..) | State::Closed { .. } => Ok(()),
+        }
+    }
+
     fn on_syn(&mut self, seq_num: u16) {
         let err = match self.endpoint {
             // If we are the accepting endpoint, then check whether the SYN is a retransmission. A
@@ -852,14 +881,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         }
     }
 
-    fn on_state(
-        &mut self,
-        seq_num: u16,
-        ack_num: u16,
-        selective_ack: Option<&SelectiveAck>,
-        delay: Duration,
-        now: Instant,
-    ) {
+    fn on_state(&mut self, seq_num: u16, ack_num: u16) {
         match &mut self.state {
             State::Connecting(connected) => match self.endpoint {
                 // If the STATE acknowledges our SYN, then mark the connection established.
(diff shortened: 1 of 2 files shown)


src/sent.rs
```diff
@@ -153,10 +169,48 @@ impl SentPackets {
         self.congestion_ctrl.on_transmit(seq_num, transmit).unwrap();
     }
 
+    /// Error if `ack_num` is not within the sequence number range of the sent packets.
+    pub fn on_ack(
```
carver, line 173: TODO: document the panic here, too (because it calls `on_full_ack`)

carver:
A more complete solution would probably be to move the ack handling into `on_packet()`. For example, the `ack_num` on a reverse-direction DATA packet is also currently ignored in the code. Putting it all in one place would reduce the code duplication, as well as slim down the number of function arguments to `on_state()` and `on_fin()`.

carver:
I'll rebase this on #109 when that's merged

src/conn.rs
```diff
@@ -762,23 +761,24 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
...
+
+        // Mark sent packets as acked, using the received ack_num
+        match packet.packet_type() {
+            PacketType::Syn => {}
+            PacketType::State | PacketType::Data | PacketType::Fin => {
+                let delay = Duration::from_micros(packet.ts_diff_micros().into());
+                if let Err(err) =
+                    self.received_ack(packet.ack_num(), packet.selective_ack(), delay, now)
+                {
+                    tracing::warn!(%err, ?packet, "ack does not correspond to known seq_num");
```
jacobkaufmann, line 778: we probably don't want to log the entire packet here

src/conn.rs
```diff
@@ -1255,14 +1252,14 @@ mod test {
...
-        conn.on_state(seq_num, syn, None, DELAY, now);
+        conn.on_state(seq_num, syn);
 
         assert!(std::matches!(conn.state, State::Established { .. }));
     }
 
     #[test]
-    fn on_state_established_invalid_ack_num() {
+    /// If a received ack_num is not a known sequence number, then reset the connection.
+    fn resets_if_ack_num_is_invalid() {
```
jacobkaufmann, line 1262:
```suggestion
    fn on_packet_invalid_ack_num() {
```

src/sent.rs
```diff
@@ -153,10 +169,49 @@ impl SentPackets {
...
+        } else {
+            Err(SentPacketsError::InvalidAckNum)
+        }
+    }
+
     /// # Panics
     ///
     /// Panics if `ack_num` does not correspond to a previously sent packet.
-    pub fn on_ack(
+    fn on_full_ack(
```
jacobkaufmann, line 214: I don't really understand what makes the ack "full" here

src/sent.rs
```diff
@@ -153,10 +169,49 @@ impl SentPackets {
...
+    /// Returns Error, if `ack_num` is not within the sequence number range of the sent packets.
+    pub fn on_ack(
+        &mut self,
+        ack_num: u16,
+        selective_ack: Option<&SelectiveAck>,
+        delay: Duration,
+        now: Instant,
+    ) -> Result<(CircularRangeInclusive, Vec<u16>), SentPacketsError> {
+        let range = self.seq_num_range();
+        if range.contains(ack_num) {
```
jacobkaufmann, line 183: IMO cleaner to short-circuit for error and then remaining code does not need to be indented

```suggestion
        if !range.contains(ack_num) {
            return Err(SentPacketsError::InvalidAckNum);
        }
        // remaining code here...
```

src/conn.rs
```diff
@@ -830,6 +830,36 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         }
     }
 
+    fn received_ack(
```
jacobkaufmann, line 833: would prefer something like `on_ack` or `process_ack` here

src/sent.rs
```diff
@@ -153,10 +169,49 @@ impl SentPackets {
...
+    /// Handle an ACK for a packet with sequence number `ack_num`, and an optional selective_ack.
+    ///
+    /// Returns Error, if `ack_num` is not within the sequence number range of the sent packets.
+    pub fn on_ack(
+        &mut self,
+        ack_num: u16,
+        selective_ack: Option<&SelectiveAck>,
+        delay: Duration,
+        now: Instant,
+    ) -> Result<(CircularRangeInclusive, Vec<u16>), SentPacketsError> {
```
jacobkaufmann, line 181: is there a reason to separate the packets acknowledged by the `ack_num` and those acknowledged by the selective ACK?

src/conn.rs
```diff
@@ -1291,21 +1288,19 @@ mod test {
...
 
-        let now = Instant::now();
-        conn.on_state(
-            syn.wrapping_add(1),
-            syn_ack.wrapping_add(2),
-            None,
-            DELAY,
-            now,
-        );
+        let delay = Duration::from_millis(100); // Dummy delay, value is irrelevant
```
jacobkaufmann, line 1291: nit: don't use "side" comments

src/sent.rs
```diff
@@ -153,10 +169,49 @@ impl SentPackets {
...
+            // the range beyond `ack_num`.
+            let full_acked = CircularRangeInclusive::new(range.start(), ack_num);
+
+            let selected_acks = if let Some(selective_ack) = selective_ack {
+                selective_ack
+                    .acked()
+                    .iter()
+                    .enumerate()
+                    .filter(|(_, &acked)| acked)
+                    .map(|(i, _)| ack_num.wrapping_add(i as u16).wrapping_add(2))
```
jacobkaufmann, line 199:
```suggestion
                    .map(|(i, _)| ack_num.wrapping_add(2).wrapping_add(i as u16))
```

jacobkaufmann (commented):
took an initial pass, left a couple of questions as well as some other minor comments

src/sent.rs
carver: Yeah, definitely open to suggestions. "full" is trying to reference the `ack_nr` acknowledgments, as opposed to "selective" ack.

src/sent.rs
carver: Mainly that the range is an efficient way to refer to many acknowledged sequence numbers, but the selective acks can't generally be encoded that way.

Maybe we could chain together an efficient range iterator, with an arbitrary vector of u16's, and return the iterator. Though I'm not really sure how to implement that in Rust, yet.

src/sent.rs
carver: What is helpful about reordering it this way?

src/sent.rs
carver: Maybe `on_ack_num`?

src/conn.rs
carver: Yeah, the debug is custom, and already skips the fields that are super wordy. It's been helpful once already!

PR #109, opened 2023-08-26
refactor: don't wrap empty fin data in Option

carver:
It's treated the same as non-empty data, so get rid of all the branch handling.


carver:
Verbal 👍🏻 from Jacob :)

PR #110, opened 2023-09-04
feat: attempt connections a bit more aggressively

carver:
- By default, try up to 6 times instead of 5
- Retry after increasing timeout by 50% instead of 100%, which helps speed up the 6th attempt.
- Increase default connection timeout to 60s, to support 6th attempt

This is the last configuration change required to reliably pass the CI test with 2,000 concurrent connections.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -127,12 +127,12 @@ pub struct ConnectionConfig {
     pub target_delay: Duration,
 }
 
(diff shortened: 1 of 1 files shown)


carver:
Very minor change. I want to get this and the tests in before the next release, so I'll just merge it. Obviously will follow up if something looks wonky.

PR #117, opened 2023-10-08
chore: add peer info to debug starting log

KolbyML:
A lot of the time I am debugging unless a connection fails there are no logs on information about the peer. This makes things hard because without this information I have no clue which uTP implementation I am looking at.


carver:
Looks like we'll probably have to reduce this concurrency number a bit to get consistently passing in CI

carver (approved):
Yeah, makes sense 👍🏻 

Let's try to get CI improved before merging

carver:
I got an error in testing at 1300, so let's just drop it to a nice round 1000. I ran 10 in a row on a Circle server without an issue. It's not foolproof, but better than status quo.

PR #119, opened 2023-12-12
POC: display packet loss on socket 1000 streams test

KolbyML:


KolbyML:
image

POC these are the packet in and out counts for both of the sockets.

I have been able to replicate packet loss on even only 50 connections. This suggestions that congestion control isn't working, as this is the exact thing it is meant to prevent against.

KolbyML:
Oh course some packet loss is to be expected, I am debugging if congestion control is working as expected currently so I am still researching what is up

PR #120, opened 2024-01-05
fix: allow Connection::eof to send empty result with one instead of two calls to Connection::process_reads()

KolbyML:
## what are the current problems 
https://github.com/ethereum/utp/blob/83d256faafa493b500f4378883bc18f5010a142d/src/conn.rs#L590-L625
- Because of how ``pending_reads`` works it can only ever contain one read channel to send data back with. This seems like a mistake, because if the code was intentionally written with only one read per ``loop { select!() }`` in a lockstep pattern, why would this code be written using ``while let`` instead of ``if let``.
- This pattern also causes another problem if the ``recv_buf`` uses the only ``pending_reads`` channel avaliable per ``process_reads()`` call. ``if self.eof() {`` can't send back it's result, which is unidle as then it has to wait to the for the next call of ``process_reads()`` to happen assuming ``pending_reads`` is refilled with another channel which is unidle as ``read_to_eof`` could have just been returned in the last call of ``process_reads()`` if this lockstep pattern wasn't in place.

So with this PR ``read_to_eof()`` will return the data after one call instead of two calls of ``Connection::process_reads()`` assuming ``self.eof()`` returns true of course.

stream.rs
```nim
    pub async fn read_to_eof(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
        // Reserve space in the buffer to avoid expensive allocation for small reads.
        buf.reserve(2048);

        let mut n = 0;
        loop {
            let (tx, rx) = oneshot::channel();
            self.reads
                .send((buf.capacity(), tx))
                .map_err(|_| io::Error::from(io::ErrorKind::NotConnected))?;

            match rx.await {
```

conn.rs
```nim
    fn on_read(&mut self, read: Read) {
        let (len, tx) = read;
        match &mut self.state {
            State::Connecting(..) | State::Established { .. } | State::Closing { .. } => {
                self.pending_reads.push_back((len, tx))
            }
            State::Closed { err } => {
                let result = match err {
                    Some(err) => Err(io::Error::from(io::ErrorKind::from(*err))),
                    None => Ok(vec![]),
                };
                let _ = tx.send(result);
            }
        }

        self.process_reads();

        self.readable.notify_waiters();
    }
```
Here is the ``read_to_eof()`` the loop generates a channel then awaits it so only one can be processed per by ``process_reads()`` after ``on_read()`` puts the channel in ``pending_reads`` deque.

This is what I meant it was stuck in a lockstep pattern to send data back.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -111,7 +111,7 @@ impl<const N: usize> fmt::Debug for State<N> {
 }
 
 pub type Write = (Vec<u8>, oneshot::Sender<io::Result<usize>>);
-pub type Read = (usize, oneshot::Sender<io::Result<Vec<u8>>>);
+pub type Read = io::Result<Vec<u8>>;
 
 #[derive(Clone, Copy, Debug)]
 pub struct ConnectionConfig {
@@ -165,7 +165,7 @@ pub struct Connection<const N: usize, P> {
     peer_recv_window: u32,
     socket_events: mpsc::UnboundedSender<SocketEvent<P>>,
     unacked: HashMapDelay<u16, Packet>,
-    pending_reads: VecDeque<Read>,
+    reads: mpsc::UnboundedSender<Read>,
     readable: Notify,
     pending_writes: VecDeque<Write>,
     writable: Notify,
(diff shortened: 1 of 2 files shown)


KolbyML:
I am also closing this as it isn't worth while. Debugging Trin more using grafana and such I am finding problems what have more substance. For example why optimize uTP's usage of packets when the biggest problem relating to that is Discv5 TaskReq responses which are just wasted packets.

So currently we send 2 udp packets for every 1 uTP packet we send.

PR #121, opened 2024-01-06
fix: if everything but our fin was acked success there fin doesn't matter

KolbyML:
This is so clients which support one way acks close successfully.
In this case we are the data sender so if they have acked everything but our fin we will count them as successful. one way acked implementations wouldn't send us a fin so I removed that check for their fin.

With this change at least trace_gossip won't read fluffy transfers as fails and keep trying to resend the content.
This doesn't delay the recv getting the content in anyway as this is a sender's side change.

Of course perfect interop would be nice, but there is no increased delay on receiving data. The only delay is closing the connection to the sender, which can slow down the trace_gossip process, but really isn't that big of a deal

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -325,7 +325,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
 
                         let unacked: Vec<u16> = self.unacked.keys().copied().collect();
                         let finished = match self.state {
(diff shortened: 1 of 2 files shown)


KolbyML:
Closing this as after debugging other Trin issues 
image
with these changes. It looks like fluffy doesn't finish due to timeout, but due to fluffy sending us a reset request because it already closed. I totally forget that being a thing after I fixed us not handling the reset requests. But that is very interesting nontheless!!!!!!!!

PR #130, opened 2024-03-16
fix: deadlock + memory leak caused by cid generation code

KolbyML:
# Issue
In Trin we were calling `cid()` which was generating cids and putting them in a Hashset of other CIDs. The issue being this set was isolated and only used for generating cids. So if you accept content from a node enough times you would be stuck in a loop because old values would never be cleared, and there would never be a unique key which didn't exist so you would loop forever, trying to find a cid which didn't already exist. There also was no code to remove cid's from this hashset

This for one created a deadlock because the loop could never escape because a unique cid was impossible, but it also wasted tons of memory because the amount of times you inserted into the Hashset before you deadlocked was tied to the non-random variable aka the peer or in trins case the node id.

So 5 nodes could put 30000k inputs each, but only when 1 node reached Xk inputs would it deadlock.

# How was it fixed

By instead using a unique hashset just for generating unique cids and nothing else. Use the active connections hashmap to tell if a cid is already used or not. The connections hashmap already handles removing and inserting values so this fixes this problem well

### DIFF
--- a/src/cid.rs
+++ b/src/cid.rs
@@ -1,4 +1,3 @@
-use std::collections::HashSet;
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::net::SocketAddr;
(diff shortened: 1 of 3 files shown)


KolbyML:
- that is just duplication between conns hashmap and the hashset
- why maintain 2 lists for the same purpose
- that is more error prone, we wouldn't have had this because for what 2 years now. If we used one object for connections this would have been solve quite a while ago.
- wasted memory storing 2 copies of a peer instead of 1

PR #134, opened 2025-01-19
refactor!: simplify UtpSocket api

morph-dev:
Remove `accept`/`connect` in favor of `accept_with_cid`/`connect_with_cid` (which are renamed to `accept`/`connect`).

This is needed in order to fix ethereum/trin#1596, which will be done in future PR.

Also simplified tracking of accepted and incoming connections: using `HashMapDelay` instead of `HashMap+HashSetDelay` combination.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1,14 +1,12 @@
-use std::collections::hash_map::DefaultHasher;
 use std::collections::HashMap;
 use std::io;
 use std::net::SocketAddr;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-use delay_map::HashSetDelay;
+use delay_map::HashMapDelay;
 use futures::StreamExt;
 use rand::{thread_rng, Rng};
-use std::hash::{Hash, Hasher};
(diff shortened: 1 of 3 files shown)


KolbyML:
So we are dropping support for using uTP as a standard replacement for TCP?

KolbyML:
Also I would argue that this isn't a refactor as it is removing functionality, it is a `feat`

KolbyML:
https://github.com/ikatson/rqbit A Rust Bittorrent client is likely to use uTP. I think it would be great if we could collaborate on 1 codebase and make it really good.

If we removed `connect/accept` as done in this PR projects like `rqbit` wouldn't be able to use `ethereum/utp`, and instead be forced to build there own uTP or add this `connect/accept` api back.

KolbyML:
Here is an issue on uTP https://github.com/ikatson/rqbit/issues/277

morph-dev:
I didn't know initially how to preserve this functionality and fix the bug. But now I think I know a way, so I will give it a try. 
Closing this PR.

PR #135, opened 2025-01-26
refactor: use HashMapDelay for awaiting and incoming connections

morph-dev:
Currently we are using HashMap+HashSetDelay combination for tracking accepted and incoming connection.

This PR simplifies it by using HashMapDelay instead.


KolbyML (approved):
:shipit: looks clean

PR #136, opened 2025-01-28
feat!: distinguish peer and peer_id on api level

morph-dev:
This PR allows `ConnectionPeer` to expose `PeerId` type and merge function.

This allows utp library to not require Peer object itself, but only PeerId object.

If multiple peer objects are provided for the same connection, they will be merged together and passed during `AsyncUdpSocket::send_to`.

The https://github.com/ethereum/trin/pull/1652 exposes how this works within trin client.

Note: I updated circleci rust version to 1.81.0 because old version was complaining about "Associated type bounds" syntax [`P: ConnectionPeer<Id: Unpin> + Unpin + 'static`](https://github.com/ethereum/utp/blob/16087be7139fc9827741cc4f23fdca061252bf50/src/socket.rs#L63)

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -8,10 +8,12 @@ use delay_map::HashMapDelay;
 use futures::StreamExt;
 use tokio::sync::{mpsc, oneshot, Notify};
 
-use crate::cid::{ConnectionId, ConnectionPeer};
+use crate::cid::ConnectionId;
 use crate::congestion;
 use crate::event::{SocketEvent, StreamEvent};
 use crate::packet::{Packet, PacketBuilder, PacketType, SelectiveAck};
+use crate::peer::ConnectionPeer;
+use crate::peer::Peer;
 use crate::recv::ReceiveBuffer;
 use crate::send::SendBuffer;
 use crate::sent::{SentPackets, SentPacketsError};
@@ -167,9 +169,10 @@ impl From<ConnectionConfig> for congestion::Config {
     }
 }
 
-pub struct Connection<const N: usize, P> {
+pub struct Connection<const N: usize, P: ConnectionPeer> {
     state: State<N>,
-    cid: ConnectionId<P>,
+    cid: ConnectionId<P::Id>,
+    peer: Peer<P>,
     config: ConnectionConfig,
     endpoint: Endpoint,
     peer_ts_diff: Duration,
@@ -185,7 +188,8 @@ pub struct Connection<const N: usize, P> {
 
 impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
     pub fn new(
-        cid: ConnectionId<P>,
+        cid: ConnectionId<P::Id>,
+        peer: Peer<P>,
         config: ConnectionConfig,
         syn: Option<Packet>,
         connected: oneshot::Sender<io::Result<()>>,
@@ -212,6 +216,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         Self {
             state: State::Connecting(Some(connected)),
             cid,
+            peer,
             config,
             endpoint,
             peer_ts_diff,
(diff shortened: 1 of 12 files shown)


src/peer.rs
```diff
@@ -0,0 +1,81 @@
...
+pub trait ConnectionPeer: Debug + Clone + Send + Sync {
+    type Id: Debug + Clone + PartialEq + Eq + Hash + Send + Sync;
+
+    /// Returns peer's id
+    fn id(&self) -> Self::Id;
+
+    /// Merge the given object into `Self` whilst consuming it.
+    ///
+    /// Should panic if ids are not matching.
+    fn merge(&mut self, other: Self);
```
KolbyML, line 15: This is kinda of confusing, merge normally means to combine 2 things, but this is more liking updating the internal copy to the latest copy seen. So when I see `merge` being used, I am confused, also the doc confuses me. Forcing me to read read the code and how it is used.

Why not use like `update_peer()` or something make its action a bit more described.

This isn't really a blocker, I am naming this function `merge` just really confuses me

KolbyML (approved):
:shipit: PR looks good, I really don't like the name `merge()` as I find it very confusing, but it isn't a blocker

src/peer.rs
morph-dev: I agree that merge is not the best name. But update_peer also is not. Let me explain.

Since this is part of the trait, each implementation can choose what it means for that one.
Some (e.g. `SocketAddr`) will choose to do nothing. Others (e.g. `Enr` that trin uses) will pick whichever is the latest (i.e. using Enr's seq field).

So while `merge` isn't the best and it's confusing, I would say that `update_peer` is also not correct (as with Enr case, it's not always updating).

Happy to consider any other name and to add clarification to the documentation (but I would avoid `update`).

KolbyML: Then maybe compare_and_swap makes more sense https://en.wikipedia.org/wiki/Compare-and-swap

This would fit both cases no? as for the `SockerAddr` we are comparing but since they are already the same we aren't swapping

for the `Enr` case we are literally doing a `compare_and_swap` operation

morph-dev: Discussed offline. I ended up using "consolidate" and provided more detailed explanation.

I also slightly changed the signature. but only for `ConnectionPeer`, not for the `Peer`. If you feel strongly about both having the same signature, let me know (in which case, I would probably use signature from `ConnectionPeer`).

KolbyML (approved):
:shipit: looks good

src/conn.rs
```diff
@@ -232,15 +237,15 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         mut writes: mpsc::UnboundedReceiver<Write>,
         mut shutdown: oneshot::Receiver<()>,
     ) -> io::Result<()> {
-        tracing::debug!("uTP conn starting... {:?}", self.cid.peer);
+        tracing::debug!("uTP conn starting... {:?}", self.cid.peer_id);
```
carver, line 240: Is it still valuable to see the `ConnectionPeer` embedded in the `Peer` here? I imagine it could be, sometimes
```suggestion
        tracing::debug!("uTP conn starting... {:?}", self.peer);
```

carver (approved):
LGTM

I like the name `consolidate()`.

Just added a couple thoughts, nothing blocking or opinionated.

src/peer.rs
```diff
@@ -0,0 +1,96 @@
...
+pub trait ConnectionPeer: Debug + Clone + Send + Sync {
+    type Id: Debug + Clone + PartialEq + Eq + Hash + Send + Sync;
+
+    /// Returns peer's id
+    fn id(&self) -> Self::Id;
+
+    /// Consolidates two peers into one.
+    ///
+    /// It's possible that we have two instances that represent the same peer (equal `peer_id`),
+    /// and we need to consolidate them into one. This can happen when [Peer]-s passed with
```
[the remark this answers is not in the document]
morph-dev: I'm not sure if you are unsure about `[Peer]` or the `-s` part. First one is link to the struct definition (see more [here](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html)) and `-s` one is my attempt on making a plural of a term that is linkified.

PR #153, opened 2025-03-25
fix: clipped data bug after SYN->STATE packet drop

### LINKED ISSUE #152: First part of transfer is ignored by recipient if the STATE response to SYN is dropped
Message flow that leads to the problem:

1. recipient initiates connection with SYN
2. sender sends STATE packet to acknowledge connection
3. [STATE packet gets lost for whatever reason?]
4. sender starts sending DATA packets
5. recipient re-attempts to initiate connection with SYN
6. sender repeats STATE packet to acknowledge connection, but with a later sequence number (subsequent to the DATA packets)
7. recipient starts receiving the content from the later sequence number, ignoring all data sent with a sequence number lower than the 2nd STATE packet

Bug first reported in https://github.com/ethereum/trin/issues/1610

See the whole process of identifying the issue above.

In short, this patch seems to fix it:
```diff
diff --git a/src/conn.rs b/src/conn.rs
index d2336e0..23bc8e0 100644
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -184,6 +184,7 @@ pub struct Connection<const N: usize, P: ConnectionPeer> {
     pending_writes: VecDeque<QueuedWrite>,
     writable: Notify,
     latest_timeout: Option<Instant>,
+    syn_ack: Option<Packet>,
 }
 
 impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
@@ -228,6 +229,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
             pending_writes: VecDeque::new(),
             writable: Notify::new(),
             latest_timeout: None,
+            syn_ack: None,
         }
     }
 
@@ -254,6 +256,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
             }
             Endpoint::Acceptor((syn, syn_ack)) => {
                 let state = self.state_packet().unwrap();
+                self.syn_ack = Some(state.clone());
                 self.socket_events
                     .send(SocketEvent::Outgoing((state, self.peer.clone())))
                     .unwrap();
@@ -787,7 +790,16 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         // NOTE: We should probably move this so that multiple incoming packets can be processed
         // before we send a STATE.
         match packet.packet_type() {
-            PacketType::Syn | PacketType::Fin | PacketType::Data => {
+            PacketType::Syn => {
+                if let Some(state) = &self.syn_ack {
+                    let event = SocketEvent::Outgoing((state.clone(), self.peer.clone()));
+                    if self.socket_events.send(event).is_err() {
+                        tracing::warn!("Cannot re-transmit syn ack packet: socket closed channel");
+                        return;
+                    }
+                }
+            }
+            PacketType::Fin | PacketType::Data => {
                 if let Some(state) = self.state_packet() {
                     let event = SocketEvent::Outgoing((state, self.peer.clone()));
                     if self.socket_events.send(event).is_err() {
```

TODO:
- [ ] add a test of dropping the SYN-acknowledging packet
- [ ] Is the above patch the best solution? Should we rebuild the state packet but selectively hold back the sequence number instead?
- [ ] Are there other problems from having the syn-ack dropped? (what if data is flowing both ways?)
- [ ] Can we incorporate a related test into portal-hive?

carver:
Fix #152 

The quick background is found in the link above. I expect it to be easier to review each commit separately.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -184,6 +184,11 @@ pub struct Connection<const N: usize, P: ConnectionPeer> {
     pending_writes: VecDeque<QueuedWrite>,
     writable: Notify,
     latest_timeout: Option<Instant>,
+    /// Save the first STATE packet returned to an incoming connection request.
+    /// We resend the exact packet in case the first response to the SYN packet is lost,
+    /// because regenerating the STATE packet gives us the wrong sequence number, and causes the
+    /// connecting peer to ignore earlier data sent to them.
+    syn_state: Option<Packet>,
 }
 
 impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
@@ -228,6 +233,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
             pending_writes: VecDeque::new(),
             writable: Notify::new(),
             latest_timeout: None,
+            syn_state: None,
         }
     }
 
@@ -240,7 +246,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         tracing::debug!("uTP conn starting... {:?}", self.peer);
 
         // If we are the initiating endpoint, then send the SYN. If we are the accepting endpoint,
-        // then send the SYN-ACK.
+        // then send the STATE in response to the SYN.
         match self.endpoint {
             Endpoint::Initiator((syn_seq_num, ..)) => {
                 let syn = self.syn_packet(syn_seq_num);
@@ -254,6 +260,7 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
             }
             Endpoint::Acceptor((syn, syn_ack)) => {
                 let state = self.state_packet().unwrap();
+                self.syn_state = Some(state.clone());
                 self.socket_events
                     .send(SocketEvent::Outgoing((state, self.peer.clone())))
                     .unwrap();
@@ -787,7 +794,29 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         // NOTE: We should probably move this so that multiple incoming packets can be processed
         // before we send a STATE.
         match packet.packet_type() {
-            PacketType::Syn | PacketType::Fin | PacketType::Data => {
+            PacketType::Syn => {
+                if self.syn_state.is_none() {
+                    if let Some(state) = self.state_packet() {
+                        tracing::warn!("Oddly missing SYN STATE, some risk to data validity");
+                        // The syn_state is generated at the beginning of the event_loop, so should
+                        // not be None here. So regenerating the STATE packet is a defensive
+                        // measure to keep working anyway, rather than ignore repeated SYN packets.
+                        // Regenerating runs the risk of causing the remote peer to ignore earlier
+                        // data sent to them, because the generated sequence number will be higher
+                        // than any previously sent data, which causes the remote peer to ignore
+                        // it silently, but think it received all the data.
+                        self.syn_state = Some(state);
+                    }
+                }
+                if let Some(state) = &self.syn_state {
+                    let event = SocketEvent::Outgoing((state.clone(), self.peer.clone()));
+                    if self.socket_events.send(event).is_err() {
+                        tracing::warn!("Cannot re-transmit syn ack packet: socket closed channel");
+                        return;
+                    }
+                }
+            }
+            PacketType::Fin | PacketType::Data => {
                 if let Some(state) = self.state_packet() {
                     let event = SocketEvent::Outgoing((state, self.peer.clone()));
                     if self.socket_events.send(event).is_err() {
@@ -1237,6 +1266,7 @@ mod test {
             pending_writes: VecDeque::new(),
             writable: Notify::new(),
             latest_timeout: None,
+            syn_state: None,
         }
     }
 
--- a/tests/stream.rs
+++ b/tests/stream.rs
@@ -19,12 +19,13 @@ const EXPECTED_IDLE_TIMEOUT: Duration = DEFAULT_MAX_IDLE_TIMEOUT;
 async fn close_is_successful_when_write_completes() {
     let conn_config = ConnectionConfig::default();
 
-    let ((send_link, send_cid), (recv_link, recv_cid)) = testutils::build_connected_pair();
+    let ((send_socket, send_cid), (recv_socket, recv_cid)) =
+        testutils::build_manually_linked_pair();
 
-    let recv = UtpSocket::with_socket(recv_link);
+    let recv = UtpSocket::with_socket(recv_socket);
     let recv = Arc::new(recv);
 
-    let send = UtpSocket::with_socket(send_link);
+    let send = UtpSocket::with_socket(send_socket);
     let send = Arc::new(send);
 
     let recv_one = Arc::clone(&recv);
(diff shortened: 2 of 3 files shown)


carver:
The failure in CI is to show that the test correctly captures the bug scenario. Note:
```
---- test_data_valid_when_resending_syn_state_response stdout ----
thread 'test_data_valid_when_resending_syn_state_response' panicked at tests/stream.rs:347:9:
assertion `left == right` failed
  left: 6952
 right: 9000
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'test_data_valid_when_resending_syn_state_response' panicked at tests/stream.rs:355:62:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(36), "assertion `left == right` failed\n  left: 6952\n right: 9000", ...)
```
`9000 - 6952 = 2048`, the same number of missing bytes as the original bug.

carver:
This is *a* fix, but I'm not convinced it's the best fix. I'll take a fresh look tomorrow, but there's a decent chance the final PR for review will look basically the same as the current code.

src/conn.rs
```diff
@@ -184,6 +184,11 @@ pub struct Connection<const N: usize, P: ConnectionPeer> {
     pending_writes: VecDeque<QueuedWrite>,
     writable: Notify,
     latest_timeout: Option<Instant>,
+    // Save the first STATE packet returned to an incoming connection request.
+    // We resend the exact packet in case the first response to the SYN packet is lost,
+    // because regenerating the STATE packet gives us the wrong sequence number, and causes the
+    // connecting peer to ignore earlier data sent to them.
+    syn_ack: Option<Packet>,
```
KolbyML, line 191:
```suggestion
    /// Save the first STATE packet returned to an incoming connection request.
    /// We resend the exact packet in case the first response to the SYN packet is lost,
    /// because regenerating the STATE packet gives us the wrong sequence number, and causes the
    /// connecting peer to ignore earlier data sent to them.
    syn_ack: Option<Packet>,
```

This should be a doc comment

src/conn.rs
```diff
@@ -787,7 +794,16 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         // NOTE: We should probably move this so that multiple incoming packets can be processed
         // before we send a STATE.
         match packet.packet_type() {
-            PacketType::Syn | PacketType::Fin | PacketType::Data => {
+            PacketType::Syn => {
+                if let Some(state) = &self.syn_ack {
+                    let event = SocketEvent::Outgoing((state.clone(), self.peer.clone()));
```
KolbyML, line 799:
```suggestion
                if let Some(state) = self.syn_ack.take() {
                    let event = SocketEvent::Outgoing((state, self.peer.clone()));
```

If we only use this option once, couldn't we just use take? to avoid the clone?. You probably have more context if this would make sense or not, but it looking reasonable to use take instead

tests/stream.rs
```diff
@@ -263,3 +266,90 @@ async fn close_succeeds_if_only_fin_ack_dropped() {
...
 }
+
+// Test that data is delivered successfully, even if the original SYN-ACK is dropped
+//
+// At the time of writing, a bug in this scenario causes the first bytes (2048 of them) to be
+// silently lost. The recipient thinks it received everything, but is missing the first bytes of
+// the transfer. When the recipient, who started the connection, times out the original SYN, it
+// resends. The sender has already sent some data. When the bug is active, the resent STATE
+// (SYN-ACK) packet uses the sequence number after incrementing from the previously sent state
+// data. This causes the recipient to ignore all data sent previously.
```
KolbyML, line 277:
```suggestion
/// Test that data is delivered successfully, even if the original SYN-ACK is dropped
///
/// At the time of writing, a bug in this scenario causes the first bytes (2048 of them) to be
/// silently lost. The recipient thinks it received everything, but is missing the first bytes of
/// the transfer. When the recipient, who started the connection, times out the original SYN, it
/// resends. The sender has already sent some data. When the bug is active, the resent STATE
/// (SYN-ACK) packet uses the sequence number after incrementing from the previously sent state
/// data. This causes the recipient to ignore all data sent previously.
```

shouldn't this be a doc comment

tests/stream.rs
```diff
@@ -263,3 +266,90 @@ async fn close_succeeds_if_only_fin_ack_dropped() {
...
+
+    // data to send, must be longer than 2048 bytes, which are lost in the bug scenario
+    const DATA_LEN: usize = 9000;
+    let data = [0xa5; DATA_LEN];
+
+    // send data
+    let acceptor_stream_handle = tokio::spawn(async move {
+        match acceptor_stream.write(&data).await {
+            Ok(written_len) => assert_eq!(written_len, DATA_LEN),
+            Err(e) => panic!("Error sending data: {:?}", e),
```
KolbyML, line 333:
```suggestion
            Err(err) => panic!("Error sending data: {err:?}"),
```

A little nit picky, but we might as well inline this error since it is done in other logging macros, this isn't really a concern though, but might as well be consistent :grin:

KolbyML (dismissed):
:shipit: PR looks good

src/conn.rs
```diff
@@ -787,7 +794,16 @@ impl<const N: usize, P: ConnectionPeer> Connection<N, P> {
         // NOTE: We should probably move this so that multiple incoming packets can be processed
         // before we send a STATE.
         match packet.packet_type() {
-            PacketType::Syn | PacketType::Fin | PacketType::Data => {
+            PacketType::Syn => {
+                if let Some(state) = &self.syn_ack {
```
morph-dev, line 798: Is is possible for `self.syn_ack` to be None? Should be fallback to `self.state_packet()` if that's the case?

morph-dev (dismissed):
I'm not familiar with details of the utp protocol (especially error handling), so this is mostly rubber stamp approval.

src/conn.rs
morph-dev: I would say that we don't want to `take()`. Maybe this response is also not received, and we have to do it again, or something like that.

Also, if I'm not mistaken, this packet should be small, so it shouldn't be too costly to clone.

ogenev (dismissed):

src/conn.rs
carver: The thought had occurred to me as I was writing it, and then I thought maybe a private variable shouldn't be?

Just now, I searched around a bit and found [this](https://github.com/rust-lang/rust-by-example/issues/1272), which prefers doc comments for private variables. There is no explanation why, so it's entirely an appeal to authority, which is among my least favorite. But I don't have a strong argument the other way. Happy to switch 👌🏻

tests/stream.rs
carver: I don't think so. From what I can tell, the generated documentation doesn't include the test files.

src/conn.rs
carver: It would be a pretty significant bug. `self.syn_ack` is generated at the beginning of the event_loop, so it should never hit this code path without one.

But I am happy to be defensive here: so I will just log a warning and generated it if it's missing. That's better behavior than ignoring the SYN entirely I suppose.

carver:
I also noticed that `syn_ack` was used elsewhere in the code for a slightly different meaning, so I renamed to `syn_state` to avoid confusion.

src/conn.rs
carver: For what it's worth, the generated docs seem to skip over private fields. So I'm not convinced that making comments on a private element into a doc-comment has any value. See https://docs.rs/utp-rs/0.1.0-alpha.17/utp_rs/conn/struct.Connection.html :

> pub struct Connection<const N: [usize](https://doc.rust-lang.org/nightly/std/primitive.usize.html), P: > [ConnectionPeer](https://docs.rs/utp-rs/0.1.0-alpha.17/utp_rs/peer/trait.ConnectionPeer.html)> { /* private fields */ }

KolbyML: Code editors will display the doc comments for a private variable when you are trying to use it or hover over it

image

So there is value when coding

KolbyML: image

KolbyML: image

See it is nice :)

carver: Yup, that's a good reason! 👌🏻

